Issue #10, opened 2017-02-22
Remove unexpected block version warnings


aguycalled:

aguycalled:
5e31db91097c0e99c83a86383f259ef525b2e482

PR #17, opened 2017-02-26
#10 unexpected block version warning

aguycalled:


src/main.cpp
```diff
@@ -7652,7 +7652,7 @@ static bool CheckStakeKernelHashV2(CBlockIndex* pindexPrev, unsigned int nBits,
 
     // Now check if proof-of-stake hash meets target protocol
     if (CBigNum(ArithToUint256(hashProofOfStake)) > bnTarget)
-        return error("CheckStakeKernelHash() : Hash %s does not meet target %s",hashProofOfStake.GetHex(),bnTarget.GetHex());
+        return false;
```
craigmacgregor: is it safe to remove this error return?

src/miner.cpp
```diff
@@ -202,11 +202,6 @@ CBlockTemplate* BlockAssembler::CreateNewBlock(const CScript& scriptPubKeyIn, bo
     pblock->nVersion       = 0x7;
     pblocktemplate->vTxSigOpsCost[0] = WITNESS_SCALE_FACTOR * GetLegacySigOpCount(pblock->vtx[0]);
 
-    CValidationState state;
-    if (!TestBlockValidity(state, chainparams, *pblock, pindexPrev, false, false, false)) {
```
craigmacgregor: Is it also safe to remove this block validity check?

PR #18, opened 2017-02-26
#9 proof of stake validity

aguycalled:
Proof of Stake fully tested. Branch candidate to merge with master.


configure.ac
```diff
@@ -360,7 +360,7 @@ case $host in
 
      AX_CHECK_LINK_FLAG([[-Wl,-headerpad_max_install_names]], [LDFLAGS="$LDFLAGS -Wl,-headerpad_max_install_names"])
      CPPFLAGS="$CPPFLAGS -DMAC_OSX"
-     OBJCXXFLAGS="$CXXFLAGS"
+     OBJCXXFLAGS="-std=c++11 $CXXFLAGS"
```
craigmacgregor: we will also need to add the -lcurl flag somewhere here, but that is only needed on the navtech branch currently

src/main.cpp
```diff
@@ -1178,14 +1178,6 @@ bool AcceptToMemoryPoolWorker(CTxMemPool& pool, CValidationState& state, const C
     if (tx.IsCoinStake())
         return state.DoS(100, false, REJECT_INVALID, "coinstake");
 
-    // Don't relay version 2 transactions until CSV is active, and we can be
```
craigmacgregor: this is the broadcast error i was hitting right? So we are happy to just remove the check? No other impact?

src/main.cpp
```diff
@@ -7769,11 +7763,11 @@ int64_t GetProofOfStakeReward(int nHeight, int64_t nCoinAge, int64_t nFees)
     int64_t nRewardCoinYear;
     nRewardCoinYear = MAX_MINT_PROOF_OF_STAKE;
 
-    if(nHeight < 7 * DAILY_BLOCKCOUNT)
+    if(nHeight-1 < 7 * DAILY_BLOCKCOUNT)
```
craigmacgregor: looking into this, wondering why the reward is different for under 1 week

src/main.cpp
aguycalled: I checked the code overall, it is safe to remove it.

src/main.cpp
aguycalled: It is like this in NavCoin original code.

https://github.com/navcoindev/navcoin2/blob/master/src/main.cpp Line 1337

configure.ac
aguycalled: Instead of directly adding -lcurl in *C*FLAGS, I'd add a check for the library in configure.ac. Regarding this branch everything is ok.

Issue #113, opened 2017-12-18
Import masterprivkey


navtrello:

aguycalled:
Use https://github.com/aguycalled/pycoin to derive addresses from a master private key.

```
git clone https://github.com/aguycalled/pycoin
cd pycoin
make
```

This will install the utility `ku`.

Example to derive 100 keys

```
$ ku -aW -s 0H/1H/1-100H -n NAV xprv9s21ZrDAJ8E9WQDjoiedjeuw9ioJFDEIOFJDEW9IjoikfjIEWODiKPMYyiuBGJHNd6
```

antonkolesnyk:
Thank you! This is helpful. I just tried it, and one of the private keys was the private key or the address that I have used. This is great.
I have created a thread on Reddit earlier, with most of these questions. Some people are also interested to know the answers. I'll update it with the info you gave me.
Still, for the long term, I think that the best option would be to have a feature to create a wallet from a private master key, during the first launch. Right now, there are two options: to create a new wallet, or to import wallet.dat. I think if there was a third option to import a private master key - this would be ideal. And it would not be confusing, and unlikely error-prone.
The reason why this feature is still needed in my opinion is not only ease of use: right now, without wallet.dat backup, I only can create a new wallet and import a bunch of private keys from the old addresses. Now, if I transact from this new wallet, I might get funds into some addresses that were derived from the new master key, if I was not careful enough to specifically use an address that derives from the previous master key. And also it might give me the false sense of security - I might naively think that after I imported a bunch of private keys from the old addresses, and have written down the new master key, I will be able to restore all the funds, which won't be the case. So the best way for me to make sure I never loose any coins would be to write down master private key each time I reinstall wallet from scratch - and you see how quickly this might start requiring me to carry tens of private master keys instead of one.

PR #118, opened 2017-12-30
Add importmasterprivkey command

antonkolesnyk:
I need a way to import master private key, to make sure I can safely restore from paper wallet. Backing up to a USB stick is not a good option.
The work I did is pretty mechanical - it is possible that I am doing it entirely wrong. But I am happy to make changes per code review comments.

### DIFF
--- a/src/wallet/rpcdump.cpp
+++ b/src/wallet/rpcdump.cpp
@@ -687,3 +687,56 @@ UniValue dumpmasterprivkey(const UniValue& params, bool fHelp)
          return NullUniValue;
      }
  }
+
+UniValue importmasterprivkey(const UniValue& params, bool fHelp)
+{
+    if (!EnsureWalletIsAvailable(fHelp))
+        return NullUniValue;
+
+    if (fHelp || params.size() < 1 || params.size() > 2)
+        throw runtime_error(
+            "importmasterprivkey \"navcoinmasterprivkey\" ( \"label\" rescan )\n"
+            "\nAdds a master private key (as returned by dumpmasterprivkey) to your wallet.\n"
+            "\nArguments:\n"
+            "1. \"navcoinmasterprivkey\"   (string, required) The private key (see dumpmasterprivkey)\n"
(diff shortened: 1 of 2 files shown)


aguycalled:
This implementation has nothing to do with importing a master private key to the wallet. It takes the supposedly private key as the wallet password, unlocks it and rescans for new transactions which is completely unnecessary after unlocking the wallet.

Adding an extra master private key to a wallet is not possible. Every wallet has a single one from which the new keys are derived. Multi-wallet support should be implemented in order to allow multiple wallets to be loaded at the same time, and therefor having different master private keys at the same time.

Another option would be to create an external tool which would create a new wallet.dat file from a given master private key to be later imported.

Anyway, if what you need is just to restore a paper wallet key (assuming you are using the one we provide on our website), it's enough using importprivkey - the private key on the paper wallet is not considered a master key and no new addresses are derived from there.

We are more than happy to receive contributions to the repository, but I'd kindly ask you to test the code you submit before opening a Pull Request.

aguycalled:
I don't consider overwriting a wallet's private key is safe as it can compromise user's funds if wallet is not empty. I'd rather use external tools to generate a new wallet.
NavCoin wallets are HD, no need to backup the wallet regularly unless you import new keys not derived from the master private key.
Writing down an address private key on paper, moving coins to the address and removing the wallet file is considered safe, as long as there are no other different addresses than the one you backed up involved.

deo1:
I posted this in the reddit thread.

When you send funds, it creates a new address to send the leftover staking coins to (the ones that weren’t sent).

You can work around this by sending “change” funds to a custom address in the GUI, but this could be error prone and inevitably someone out there only saved the original private key and doesn’t realize they have their funds spread across multiple addresses.

Importing masterprivkey does not work in the core wallet, as this was the first thing I tried.

I’d really love to have a single offline private key from which secondary keys are derived. Is there a way to store one key for the HD wallet from which all additional addresses that get created when transacting can be derived?

aguycalled:
Please check: https://github.com/NAVCoin/navcoin-core/issues/113

PR #136, opened 2018-01-25
gui: accepts uri to sign messages

aguycalled:
Adds support for signing messages through URIs. This will allow a website to verify the identity of an address owner.

When user clicks in a URI in the format:

navcoin:http://domain/?m=random_string&a=address

The wallet will try to sign the message random_string with the address and if successful will make a request including the resulting signature to:

http://domain/?m=random_string&a=address&s=signature

### DIFF
--- a/src/qt/paymentserver.cpp
+++ b/src/qt/paymentserver.cpp
@@ -478,12 +478,12 @@ void PaymentServer::handleURIOrFile(const QString& s)
 
             QString signature = (QString::fromStdString(EncodeBase64(&vchSig[0], vchSig.size())));
 
-            QUrl fetchUrlConstructed(s.mid(NAVCOIN_IPC_PREFIX.length()) + "&s=" + QUrl::toPercentEncoding(signature));
+            QUrl fetchUrlConstructed(s.mid(NAVCOIN_IPC_PREFIX.length()));
 
             Q_EMIT message(tr("Verify address"), tr("Message signed."),
                            CClientUIInterface::ICON_INFORMATION);
 
-            sendSignature(fetchUrlConstructed);
+            sendSignature(fetchUrlConstructed, signature);
 
         }
         else if (uri.hasQueryItem("r")) // payment request URI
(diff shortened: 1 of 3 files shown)


src/qt/paymentserver.cpp
```diff
@@ -414,7 +422,71 @@ void PaymentServer::handleURIOrFile(const QString& s)
...
-        if (uri.hasQueryItem("r")) // payment request URI
+        if (uri.hasQueryItem("m") && uri.hasQueryItem("a")) // payment request URI
+        {
+            QByteArray temp;
+            temp.append(uri.queryItemValue("m"));
+            temp.append(uri.queryItemValue("a"));
+            QString decoded = QUrl::fromPercentEncoding(temp);
+            QUrl fetchUrl(decoded, QUrl::StrictMode);
+
+            if (!model)
```
matt-auckland: should we throw an error here?

src/qt/paymentserver.cpp
```diff
@@ -414,7 +422,71 @@ void PaymentServer::handleURIOrFile(const QString& s)
...
+                return;
+
+            CNavCoinAddress addr(uri.queryItemValue("a").toStdString());
+            if (!addr.IsValid())
+            {
+                Q_EMIT message(tr("Verify address"), tr("The entered address is invalid.") + QString(" ") + tr("Please check the address and try again."),
+                               CClientUIInterface::ICON_WARNING);
+                return;
+            }
+            CKeyID keyID;
```
matt-auckland: What's a key? A Public or Private key? Or something else?

src/qt/paymentserver.cpp
```diff
@@ -414,7 +422,71 @@ void PaymentServer::handleURIOrFile(const QString& s)
...
+            if (!addr.GetKeyID(keyID))
+            {
+                Q_EMIT message(tr("Verify address"), tr("The entered address does not refer to a key.") + QString(" ") + tr("Please check the address and try again."),
+                               CClientUIInterface::ICON_WARNING);
+                return;
+            }
+
+            WalletModel::UnlockContext ctx(model->requestUnlock());
```
matt-auckland: General comment, I could be blind but it doesn't look like you're checking if:
- the string to sign is too long
- the string to sign is too short (i.e. length = 0)
- the url is valid

src/qt/paymentserver.cpp
```diff
@@ -642,6 +714,15 @@ void PaymentServer::fetchRequest(const QUrl& url)
     netManager->get(netRequest);
 }
 
+void PaymentServer::sendSignature(const QUrl& url)
+{
```
matt-auckland: Do we need to error if this fails?

src/qt/paymentserver.cpp
aguycalled: This is just a sanity check to prevent referencing a null pointer. Should never happen (as model is defined on startup). Gui code is full of checks like this and it never throws an error.

src/qt/paymentserver.cpp
aguycalled: In this case, a public key.

src/qt/paymentserver.cpp
aguycalled: In case the request does not succeed, the user is notified.

craigmacgregor:
Code looks fine to me. I haven't compiled and tested it yet. I guess the only concern here is making sure that there is absolutely no way for the users private key to somehow be exposed to the request sent back to the server due to some kind malicious injection attempt. Are we happy that the input strings are correctly sanitised or otherwise unable to become an attack vector and that if an error was encountered the wallet is unable to return any data which could compromise the private key used to sign the message. From looking at the code, this all looks like it's accounted for already, but just wanted to voice the thought.

red010b37:
I had a look and from what little i could understand it seemed ok

PR #138, opened 2018-01-29
V4.1.2 cfund logic fix

aguycalled:
fixes a logic bug where expired proposals never get the expired state and sit indefinitely waiting for end of the voting period


aguycalled:
pushed a few more commits re: cfund categories for transactions on the gui

red010b37:
@aguycalled so do we have a completely new payment server for the CFund?

aguycalled:
the uri handler code is hosted in paymentserver.* but the new functionality is not related to the cfund but introduces a way to verify address ownership from websites using an specific uri scheme

Issue #147, opened 2018-02-14
Prioritise community fund proposals by paid fee


navtrello:
CFund proposals are stored using LevelDB, being the hash of the proposal the key. This causes proposals to be randomly prioritised when, for example, there are no enough funds on the fund and the consensus protocol needs to decide which proposal is the first with right to claim coins.
To solve this CCFundDB::GetProposalIndex() should sort the vector based on the paid fee of the proposal, so proposals with a higher fee are prioritised.

┆Attachments: <a href="https:&#x2F;&#x2F;github.com&#x2F;NAVCoin&#x2F;navcoin-core&#x2F;commit&#x2F;5fa006aba6023afc6131c9ab8b82e3033bd95235">cfund: prioritise proposals by fee</a>

navtrello:
➤ alex v commented:

https://github.com/NAVCoin/navcoin-core/commit/5fa006aba6023afc6131c9ab8b82e3033bd95235

navtrello:
➤ alex v commented:

waiting for the other 2 pull requests to be approved to open a new one

PR #148, opened 2018-02-16
V4.1.2 cfund prio fee

aguycalled:
CFund proposals are stored using LevelDB, being the hash of the proposal the key. This causes proposals to be randomly prioritised when, for example, there are no enough funds on the fund and the consensus protocol needs to decide which proposal is the first with right to claim coins.
To solve this CCFundDB::GetProposalIndex() should sort the vector based on the paid fee of the proposal, so proposals with a higher fee are prioritised.
#147

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -3535,7 +3535,7 @@ bool CountVotes(CValidationState& state, CBlockIndex *pindexNew, const CBlock *p
                         proposal.fState = CFund::EXPIRED;
                         fUpdate = true;
                     }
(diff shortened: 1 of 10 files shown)


matt-auckland:
This seems to have some other changes mixed in with it, is this what you were talking about the other day when you mentioned other changes showing up in the diff?

aguycalled:
Yes, that's what I was talking about. 
The only change is contained in this commit: https://github.com/NAVCoin/navcoin-core/pull/148/commits/5fa006aba6023afc6131c9ab8b82e3033bd95235

PR #151, opened 2018-02-22
V4.1.2 cfund check db

aguycalled:
If the CFund DB is removed or corrupted there'll be incongruences on the state of the proposals and differences in the consensus between nodes.

CFund DB should be rebuilt in case it is not synced up to tip of the chain.

### DIFF
--- a/src/txdb.cpp
+++ b/src/txdb.cpp
@@ -20,6 +20,7 @@ static const char DB_COINS = 'c';
 static const char DB_BLOCK_FILES = 'f';
 static const char DB_TXINDEX = 't';
 static const char DB_PROPINDEX = 'p';
+static const char DB_PROP_TIP_HEIGHT = 't';
 static const char DB_PREQINDEX = 'r';
 static const char DB_ADDRESSINDEX = 'a';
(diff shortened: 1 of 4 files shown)


aguycalled:
[Check consistency of CFund DB](https://trello.com/c/2bkblYjw/262-check-consistency-of-cfund-db)

src/txdb.cpp
```diff
@@ -268,6 +269,18 @@ bool CCFundDB::GetPaymentRequestIndex(std::vector<CFund::CPaymentRequest>&vect)
...
+
+bool CCFundDB::WriteTipHeight(int nHeight) {
+    return Write(DB_PROP_TIP_HEIGHT, nHeight);
+}
+
+int CCFundDB::ReadTipHeight() {
+    int nHeight = 0;
+    if (!Read(DB_PROP_TIP_HEIGHT, nHeight))
+        return 0;
+    return nHeight;
```
ro-savage: As far as I can tell this always turns 0. Is that was is meant to happen?

aguycalled: it would return 0 if its not able to read the chain tip height into nHeight, otherwise it would return the stored value

Issue #117, opened 2017-12-26
Witness Merkle Root with Fund Enabled


aguycalled:
Using Coinbase outputs to emit payments for approved payment requests breaks the Coinbase Witness Commitment and makes P2W* scripts not spendable.

- - -
[Elegantt data. What's this?](http://bit.ly/elegantt-for-trello-whats-this)
[](Elegantt!UNITO-UNDERSCORE!data:dont!UNITO-UNDERSCORE!delete{"ignored":false,"autoPlanned":true,"ownerId":false,"dependencies":[],"psd":"2018-01-21T15:00:00+13:00","ped":"2018-01-21T22:30:00+13:00"})

┆Attachments: <a href="https:&#x2F;&#x2F;github.com&#x2F;NAVCoin&#x2F;navcoin-core&#x2F;pull&#x2F;154">navcoin-core #154 fix: #117 ignores coinstake output when calculating witness merkle root</a>

PR #154, opened 2018-02-27
fix: #117 ignores coinstake output when calculating witness merkle root

aguycalled:
The witness commitment is calculated when a new block template is created before the coin stake is added. This can cause the witness merkle root hash to be different when later checked in a completely constructed block broadcasted to the network. This patch takes in consideration if the coin stake should be ignored, preventing the last leaf of the tree to be an empty hash.


aguycalled:
[Witness Merkle Root with Fund Enabled](https://trello.com/c/Mve86SCV/177-witness-merkle-root-with-fund-enabled)

Issue #83, opened 2017-09-03
(FEATURE) Retina / high dpi configuration


jjxtra:
If you add this to the main function, before initializing QT or doing anything else, it will allow the UI to look correct on retina / high dpi screens.

```
// Address 4K screen resolution: http://doc.qt.io/qt-5/highdpi.html
// enable automatic scaling based on the pixel density of the monitor
qputenv( "QT_AUTO_SCREEN_SCALE_FACTOR", "1" );
```

[](Elegantt!UNITO-UNDERSCORE!data:dont!UNITO-UNDERSCORE!delete{"ignored":false,"autoPlanned":true,"ownerId":false,"dependencies":[],"psd":"2017-11-21T23:44:29+13:00","ped":"2017-11-22T03:16:15+13:00"})

aguycalled:
Are you affected by this issue? Could you help us testing compiling and running branch v4.0.5? thx

jjxtra:
Sure thing, will do. As a side question, how do I move the blockchain files to another drive? My c drive is small, but I have a big d drive.

aguycalled:
Just look for the folder NavCoin4 in %APPDATA%, there you'll find two folders 'blocks' and 'chainstate' where the copy of the blockchain is stored. Move those folders (and ideally also your wallet.dat file) to the new desired location and specify it using the datadir argument when running the wallet binary.

more info: https://bitcoin.stackexchange.com/questions/9259/bitcoin-qt-setup-change-block-chain-file-location-settings-documentation

jjxtra:
I've only got Windows here and was looking for a .pro to run qmake and build and run in qt creator, but didn't see anything. Am I just missing it? If there isn't a .pro project file, is there value in creating one? Would simplify the build greatly as you could just build and run in qt creator.

aguycalled:
You can just load the project folder in QT Creator to compile and run it, that's how I do it, no need for .pro file. Anyway if you are in windows I'm not sure if you are going to be able to solve all the dependencies without using a VM.

https://github.com/NAVCoin/navcoin-core/blob/master/doc/build-windows.md

navtrello:
➤ Matt commented:

is this solved?

aguycalled:
pending testing
binaries available at https://build.nav.community/v4.1.2/

red010b37:
Released with 4.21.

PR #157, opened 2018-02-28
gui: #83 fix pixel density on hidpi screens

aguycalled:
ui showed broken on hidpi screens.
needs to be tested on various os and screen resolutions before merged.


matt-auckland:
Where can I find a build for this?

PR #158, opened 2018-02-28
net: update hardcoded nodes

aguycalled:


aguycalled:
contrib/seeds/generate-seeds.py is used to convert nodes_*.txt in a header file (the hex format you are seeing)

PR #160, opened 2018-02-28
V4.1.2 no reindex before cfund

aguycalled:
If the CFund db is corrupted or has been removed, the wallet will ask the user to reindex/resync in order to be able to rebuild the CFund DB.

This does not make any sense before the Community Fund is enabled, as it would make every mainnet user to completely resync the chain when v4.1.2 is released.

A new flag is written to the block tree reflecting the current activation state of the CFund, and this is checked on init when validating the DB integrity.


matt-auckland:
"This does not make any sense before the Community Fund is enabled, as it would make every mainnet user to completely resync the chain when v4.1.2 is released." Nice catch 😛

matt-auckland:
My phone can't deal with these large files so I'll check the code when I get to the office

src/base58.cpp
```diff
@@ -188,9 +188,6 @@ bool CBase58Data::SetString(const std::string& str)
 
 std::string CBase58Data::ToString() const
 {
-//    if(vchData.To == "0x202020")
-//      return "FUND_CONTRIBUTION";
-
```
matt-auckland: why don't we need this anymore?

aguycalled: its a ghost comment from some internal test i did i forgot to remove. we never needed it.

matt-auckland:
Ah I see

PR #163, opened 2018-02-28
V4.1.2 missing libs in distributed binaries

aguycalled:


aguycalled:
i personally tested it, and multi arch binaries were correctly built using gitian. additional testing is welcome tho.

Issue #166, opened 2018-03-12
Wallet not able to sync


aguycalled:
Sockets are closed after a few block are received.

Version affected: v4.1.2, probably v4.1.1

PR #167, opened 2018-03-12
net: ban obsolete nodes

aguycalled:
Related to https://github.com/NAVCoin/navcoin-core/issues/166

This patch bans nodes with obsolete versions instead of keeping the socket open.

Issue was solved in my own environment.

Testing and confirmation in other systems required.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -5922,12 +5922,9 @@ bool static ProcessMessage(CNode* pfrom, string strCommand, CDataStream& vRecv, 
 
         if(fObsolete)
         {
-            if(!pfrom->fWrongVersion) {
-                pfrom->fWrongVersion = true;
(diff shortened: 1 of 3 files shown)


aguycalled:
https://github.com/NAVCoin/navcoin-core/blob/master/src/main.cpp#L1942
Missbehaving() does log

matt-auckland:
Cool. How should we test this? Download the latest 4.1.2 build and run it to see if it bans older nodes?
How do we define an obsolete version?

aguycalled:
a node with protocol version < 70017 will be considered obsolete
70020 will be min. protocol version when cfund becomes active

important to test if a clean install of the wallet is able to sync with mainnet

rex-kwon:
testing now...

PR #172, opened 2018-03-16
cfund: verbalize errors rpc commands

aguycalled:
voting for invalid or duplicated proposals/payment requests, or trying to remove invalid or inexistent proposals/payment requests, throws an exception.

as per request of sakdeniz for navcoin vue


aguycalled:
Original request from sakdeniz:

sakdeniz - letzten Freitag um 10:25 Uhr
is it possible to send back a return value like "success" when the proposalvote command succeeds?
it would also be nice to return a duplicate message even if the proposal was previously voted
also it accepts invalid proposal hash, and does not throw an error like 'invalid proposal hash'
if possible, i need these modifications for vue wallet

PR #173, opened 2018-03-19
V4.1.2 navcoin tx

aguycalled:
Fixes navcoin-tx tool to create raw transactions.


aguycalled:
Description:
- Expands tx version range up to 255 (to fit tx with attached anon data + Community Fund tx).
- Adds option to set the transaction time (navcoin's peculiarity).
- Shows the value of time when dumping the tx in JSON format.
- Original tool only allowed to sign inputs. A 4th field is added to command IN to allow injection of arbitrary data in the scriptSig field, to ease testing new features at protocol level.

PR #177, opened 2018-03-21
cfund: memory optimization

aguycalled:
Moves declaration of some variables out of the loop, to prevent multiple unnecessary allocations and destructions of the objects thus optimising the memory usage.


src/main.cpp
```diff
@@ -3477,6 +3477,7 @@ bool static ConnectTip(CValidationState& state, const CChainParams& chainparams,
```
matt-auckland: Should we update the copyright at the top of the file? 2017-2018?

aguycalled: go for it

PR #192, opened 2018-04-12
V4.1.2 clocksync

aguycalled:
this pull request introduces mandatory clock sync against a ntp server for every node on launch.

there are some security concerns around the ntp protocol, being it susceptible to mitm attacks as packets aren’t signed for verification.
to overcome this, it’s required to check clock against a minimum number of servers (3 by default).
the number of measures from different servers is required to be odd, then later we calculate the clock drift as the median of all the different measured values.
a list of public, trusted, ntp nodes is hardcoded at utiltime.h.
syncing against an external trusted clock does not protect from a malicious attacker artificially modifying its clock during runtime introducing wrong timestamps on the network.
to overcome this, i propose reducing the tolerated drift in the future of a block timestamp to 60 seconds and to completely forbid blocks in the past.
those consensus changes would be activated through a soft fork, signaled by version bit number 8 of staked blocks.

other related changes:

- new log category “ntp”
- new multi argument -ntpserver= to allow user to manually specify servers
- new argument -ntpminmeasures= to allow user to specify the req. min. of measures
- new argument -ntptimeout= to allow user to specify how many seconds should we wait for the response of a ntp server


aguycalled:
added a check to detect if system clock is changed. 
wallet will then try to sync using ntp servers and terminate if fails.

ro-savage:
Will this do a check every time you attempt to mint a block? Or is it only when the daemon starts?

src/ntpclient.cpp
```diff
@@ -0,0 +1,195 @@
...
+
+            // We push if its the first entry
+            if(vResults.size() == 0)
+            {
+                vResults.push_back(nClockDrift);
+
+                string sign = "";
+                if(nClockDrift > 0)
+                    sign = "+";
+                else if (nClockDrift < 0)
```
craigmacgregor: can nClockDrift == 0 ? if so, the sign variable won't be defined with this logic.

src/ntpclient.cpp
```diff
@@ -0,0 +1,195 @@
...
+                    sign = "+";
+                else if (nPrevMeasure < 0)
+                    sign = "-";
+
+                sReport += sPrevServer + "[" + sign + to_string(nPrevMeasure) + "sec.] ";
+
+                nPrevMeasure = -1;
+                sPrevServer = "";
+            }
+            else
```
craigmacgregor: Would this logic be more readable if it were to work on a modulus rather than setting and unsetting the nPrevMeasure variable to -1 ? It took me a while to understand what was going in here and what happened during each cycle of the for loop.

```
if (i == 0)
    push_back(first_measure)
else if (i % 2 == 1)
    push_back(previous_measure)
    push_back(current_measure)
else 
    previous_measure = current_measure
```

src/utiltime.cpp
```diff
@@ -9,22 +9,42 @@
...
+int64_t GetTimeNow()
+{
+    if (nMockTime) return nMockTime;
+
     time_t now = time(NULL);
     assert(now > 0);
     return now;
 }
 
+int64_t GetSteadyTime()
```
craigmacgregor: what exactly is steadyTime?

http://en.cppreference.com/w/cpp/chrono/steady_clock

src/ntpclient.cpp
aguycalled: sign is "" if nClockDrift is 0.

sign is just the prefix (string) to prepend to the drift when shown on the log file

src/utiltime.cpp
aguycalled: as the link describes, is a clock that is not affected by time zone changes, cannot be regressed and every tick corresponds to a physical tick. in some systems is used to measure time since boot. this is the context I'm using it:
- on every cycle of a time interval t, 2 time measures are taken, using the steady and the normal clock each of them.
- on every cycle >0 of a time interval t, the elapsed time is calculated by substracting the value of the previous sample.
- if the elapsed time of the clocks differs between them in a value higher than a predefined threshold (1000ms in the current implementation) it's safe to assume the local clock has been artificially changed so we need to resync against ntp servers before minting a block.

aguycalled:
@craigmacgregor NtpClockSync() tries to sync clock using external trusted sources (NTP servers). Daemon is only shut down if this process fails.
Only consensus change is to limit block time to be in the range (now, now+60]

src/ntpclient.cpp
aguycalled: Done -> 5840155a7f6dfc8f6f84e34424f16e501fd62c67

src/utiltime.cpp
craigmacgregor: cool, thanks for the explanation :)

craigmacgregor:
- Setup our own NTP servers
- Add more NTP servers that are globally distributed
- Document how to override the NTP check if all the servers are down
- Document a testing plan for the testnet
- Setup a developer testnet
- Replicate the current issue on the testnet
- Run through the testing plan on the testnet

aguycalled:
added a new argument -maxtimeoffset= to set the max tolerated clock drift for peers.
also removed support to readjust clock based on data from other peers.

PR #199, opened 2018-04-23
Wallet support for bootstrapping

aguycalled:
This PR adds support for bootstrapping the blockchain directly from the wallet.

A new argument (-bootstrap) is handled on initialisation to specify an URL from where a copy of the blockchain in TAR format will be downloaded and extracted in the data folder.

Adds too in the GUI a submenu entry under FILE to do this operation graphically.


red010b37:
Sorry - again don't want to release this in 4.2.0, but could be its own patch release

src/init.cpp
```diff
@@ -800,6 +831,63 @@ void InitLogging()
  */
 bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
 {
+    // ********************************************************* Step 0: download bootstrap
```
red010b37: Does this mean it runs on every start? as its part of the App Init

src/init.cpp
```diff
@@ -800,6 +831,63 @@ void InitLogging()
...
+                        boost::filesystem::remove_all(sDownload);
+                        if(!fOk)
+                            return InitError("[BOOTSTRAP] Could not extract data.");
+                    }
+                } else {
+                    return InitError("[BOOTSTRAP] Failed!");
+                }
+            }
+        } else {
+            return InitError("[BOOTSTRAP] Failed!");
```
red010b37: Should this be more descriptive, as it's not that the bootstrap failed its the lib doesn't exist? so we exit. Thin it should say, Failed as curl is not found

src/init.cpp
```diff
@@ -800,6 +831,63 @@ void InitLogging()
...
+    if(sBootstrap != "")
+    {
+        RemoveConfigFile("bootstrap");
+        uiInterface.InitMessage(strprintf("[BOOTSTRAP] Trying to download from %s", sBootstrap.c_str()));
+        CURL *curl;
+        FILE *fp, *a;
+        CURLcode res;
+
+        curl = curl_easy_init();
+        if (curl) {
```
red010b37: is curl os specific. e.g you need it installed on your machine, or is packaged in the app? - just asking

src/init.cpp
aguycalled: yes

src/init.cpp
aguycalled: the library is linked, curl is already used in other parts of navcoin-core

src/init.cpp
aguycalled: if the lib does not exist, the code won't compile.
curl_easy_init( ) is NULL when "something went wrong" as per Curl official documentation (https://curl.haxx.se/libcurl/c/curl_easy_init.html) - hence the lack of definition in the warning text

src/init.cpp
red010b37: thanks :)

src/init.cpp
```diff
@@ -799,6 +830,63 @@ void InitLogging()
  */
 bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
 {
+    // ********************************************************* Step 0: download bootstrap
+    std::string sBootstrap = GetArg("-bootstrap","");
+
+    if(sBootstrap != "")
```
red010b37: I think this should be split out to its own function

red010b37: Thoughts?

PR #207, opened 2018-05-09
start community fund accumulation

aguycalled:
This PR introduces the changes necessary to start the accumulation of coins in the Community Fund.

A new Soft Fork signaled by version bit 7 will trigger:
- reduction of the staking rewards to 4%
- inclusion of an additional output in the coinstake tx contributing 0.25NAV to the Community Fund
- consensus validation of the previous rule


src/main.cpp
```diff
@@ -4309,6 +4315,13 @@ bool IsCommunityFundEnabled(const CBlockIndex* pindexPrev, const Consensus::Para
     return (VersionBitsState(pindexPrev, params, Consensus::DEPLOYMENT_COMMUNITYFUND, versionbitscache) == THRESHOLD_ACTIVE);
 }
 
+bool IsCommunityFundAccumulationEnabled(const CBlockIndex* pindexPrev, const Consensus::Params& params, bool fStrict)
+{
```
red010b37: I think fStrict should have a default, which i think should be true

src/main.cpp
```diff
@@ -4309,6 +4315,13 @@ bool IsCommunityFundEnabled(const CBlockIndex* pindexPrev, const Consensus::Para
...
+    LOCK(cs_main);
+    return (IsCommunityFundEnabled(pindexPrev, params) && !fStrict) ||
+          (VersionBitsState(pindexPrev, params, Consensus::DEPLOYMENT_COMMUNITYFUND_ACCUMULATION, versionbitscache) == THRESHOLD_ACTIVE);
```
red010b37: Also why do we have fStrict? it seems to be an exit for when we are connecting block. but wouldn't it be better to check it each time? - or is it a performance issue?

Just tying to understand?

src/main.cpp
```diff
@@ -8322,7 +8339,7 @@ int64_t GetProofOfStakeReward(int nHeight, int64_t nCoinAge, int64_t nFees, CBlo
         nRewardCoinYear = 0.5 * MAX_MINT_PROOF_OF_STAKE;
     else if(nHeight-1 < (730 * Params().GetConsensus().nDailyBlockCount))
         nRewardCoinYear = 0.5 * MAX_MINT_PROOF_OF_STAKE;
-    else if(IsCommunityFundEnabled(pindexPrev, Params().GetConsensus()))
+    else if(IsCommunityFundAccumulationEnabled(pindexPrev, Params().GetConsensus(), false))
```
red010b37: Same as above true - false, think one way should be default or we check each time

src/main.cpp
aguycalled: when fStrict is true, IsCommunityFundAccumulationEnabled checks only for DEPLOYMENT_COMMUNITYFUND_ACCUMULATION soft fork.
when false checks for this _OR_ DEPLOYMENT_COMMUNITYFUND

PR #213, opened 2018-05-18
V4.2.0 open alias

aguycalled:
This PR completes the implementation of the open alias protocol in the NavCoin Core client.

OpenAlias is an standard created by the Monero Core project used in the Monero wallet and some other clients like Bitcoin Electrum which allows to translate email addresses into cryptocurrency addresses using custom TXT entries in the DNS records of a domain.

More details about the specification can be found at www.openalias.org

Part of the code is based in the original Monero implementation. Main differences are:
- Prefix of the TXT entry is required to be set to **oa1:nav**
- The only parsed parameter is recipient_address
- NavCoin **enforces** by default the use of DNSSEC
- Originally Unbound was used to resolve the entry. This created issues when linking the library in some systems, so alternatively support for using Google's DNS over HTTPS service has been implemented. Future implementation of libunbound+dnscrypt is highly recommended to reduce dependence on 3rd parties

Other modifications in the NavCoin client:
- New RPC command **resolveopenalias** resolves an OpenAlias address to a NavCoin address
- Added support for sending to OpenAlias addresses in the GUI, when parsing URIs and the RPC commands **validateaddress** and **sendtoaddress**
- New argument **-requirednssec** to set whether DNSSEC validation is required to resolve openalias addresses (true by default).
- Update copyright notice

### DIFF
--- a/src/utils/dns_utils.cpp
+++ b/src/utils/dns_utils.cpp
@@ -26,8 +26,6 @@
 // STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
 // THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-#ifdef HAVE_UNBOUND
-
 #include "utils/dns_utils.h"
 #include "util.h"
 
@@ -40,11 +38,27 @@
 #include <stdlib.h>
 #include <random>
 #include <sstream>
+#ifdef HAVE_UNBOUND
 #include <unbound.h>
+#else
+#include <curl/curl.h>
+#include <univalue.h>
+#endif
 
 namespace bf = boost::filesystem;
 
 static boost::mutex instance_lock;
+
+size_t write_data(
+        const char* in,
+        std::size_t size,
+        std::size_t num,
+        std::string* out)
+{
+    const std::size_t totalBytes(size * num);
+    out->append(in, totalBytes);
+    return totalBytes;
+}
 
 namespace
 {
@@ -54,13 +68,15 @@ namespace
  * the unbound library.
  */
 
+#ifdef HAVE_UNBOUND
 /** return the built in root DS trust anchor */
 static const char*
 get_builtin_ds(void)
 {
-  return
-". IN DS 19036 8 2 49AAC11D7B6F6446702E54A1607371607A1A41855200FD2CE1CDDE32F24E8FB5\n";
+    return
+            ". IN DS 19036 8 2 49AAC11D7B6F6446702E54A1607371607A1A41855200FD2CE1CDDE32F24E8FB5\n";
 }
+#endif
 
 /************************************************************
  ************************************************************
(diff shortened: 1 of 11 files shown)


red010b37:
Don't want to add any more to v 4.2.0-rc, this should be its own minor release

src/utils/dns_utils.cpp
```diff
@@ -54,13 +70,15 @@ namespace
  * the unbound library.
  */
 
+#ifdef HAVE_UNBOUND
```
red010b37: what does HAVE_UNBOUND relate to in this context?

src/utils/dns_utils.cpp
```diff
@@ -166,348 +191,406 @@ class string_copy {
...
+     * NAV address in the correct format, returns an empty string. <dnssec_valid>
+     * will be set true or false according to whether or not the DNS query passes
+     * DNSSEC validation.
+     *
+     * @param url the url to look up
+     * @param dnssec_valid return-by-reference for DNSSEC status of query
+     *
+     * @return a navcoin address (as a string) or an empty string
+     */
+    std::vector<std::string> addresses_from_url(const std::string& url, bool& dnssec_valid)
```
red010b37: Should we have a test for this?

src/utils/dns_utils.cpp
```diff
@@ -166,348 +191,406 @@ class string_copy {
...
+        curl_easy_setopt(curl, CURLOPT_URL, serverURL.c_str());
+        curl_easy_setopt(curl, CURLOPT_TIMEOUT, 5);
+        curl_easy_setopt(curl, CURLOPT_WRITEFUNCTION, write_data);
+        curl_easy_setopt(curl, CURLOPT_WRITEDATA, &readBuffer);
+        res = curl_easy_perform(curl);
+        curl_easy_getinfo(curl, CURLINFO_RESPONSE_CODE, &httpCode);
+        curl_easy_cleanup(curl);
+
+        if (httpCode == 200) {
+            UniValue request;
```
red010b37: There seem to be a lot of ways to get addresses. Is this correct? as we seem to be just calling https://dns.google.com/resolve

What is the common path through here?

red010b37: or are we just returning an empty vector on error?

src/utils/dns_utils.cpp
```diff
@@ -166,348 +191,406 @@ class string_copy {
...
-  {
-    const std::string &url = dns_urls[cur_index];
-    if (!avail[cur_index])
+    namespace
     {
-      records[cur_index].clear();
-      LogPrintf("DNSSEC not available for checkpoint update at URL: %s, skipping.", url);
-    }
-    if (!valid[cur_index])
+    bool dns_records_match(const std::vector<std::string>& a, const std::vector<std::string>& b)
```
red010b37: and a test for this as double for loops can be a pain

src/utils/dns_utils.cpp
aguycalled: if the library unbound is present we will use it to resolve the dns entry instead of dns over https

src/utils/dns_utils.cpp
aguycalled: exactly, an empty vector is returned when no result is found (whatever the reason for this is)

PR #219, opened 2018-06-06
4.2.1 hotfix

aguycalled:
# NavCoin v4.2.1 Release Notes

### Hot Fix:

- The wallet will try to synchronize clock using a NTP Server every 30 seconds on launch until it succeds instead of just checking once and exiting if it fail
- Change the buffer size of the socket reading from the NTP Server to 32 bit, fixing some compatibiity issues with 32 bit systems.


src/init.cpp
```diff
@@ -1112,21 +1112,45 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
...
+                }
+
+                strMiscWarning = sMsg;
+                AlertNotify(strMiscWarning);
+                LogPrintf(strMiscWarning.c_str());
+
+                uiInterface.InitMessage(_(strprintf("Synchronizing clock attempt %i...", nWarningCounter+1).c_str()));
+
+                nWarningCounter++;
+
```
matt-auckland: Why do we show a string that is `nWarningCounter + 1` and then after that do `nWarningCounter++` ?
Doesn't it make more sense to increment it and then log it?

craigmacgregor: for readability and future proofing it's best to increment at the beginning or end of a statement rather than in the middle of it. I'm okay with adding the +1 in the log output.

matt-auckland: That makes sense. I'm happy to approve this then, was mostly curious about the why

PR #224, opened 2018-06-12
Community Fund DB

aguycalled:
This PR moves back the Community Fund data together with the Chain State DB.

### DIFF
--- a/src/consensus/cfund.cpp
+++ b/src/consensus/cfund.cpp
@@ -66,7 +66,7 @@ bool CFund::FindProposal(uint256 prophash, CFund::CProposal &proposal)
     }
 
     CFund::CProposal temp;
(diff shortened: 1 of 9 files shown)


aguycalled:
no, i need this before implementing versioning for proposals so i can add new parameters needed for those features without breaking backwards compatibility

aguycalled:
Updated with new changes:
- Adds support for Proposal and Payment Request Versioning, allowing to expand their properties in the future without breaking backwards compatibility of the data base.
- Duration of the voting for Proposals and Payment Requests is limited to a predefined amount of full voting cycles.
- Payment requests can still be voted after the deadline date of the parent proposal if the voting started before of the expiration of the proposal.
- Deadline of the proposals is not set relatively to the moment of acceptance of the proposal.

PR #241, opened 2018-07-10
add support to ban nodes based on broadcasted version

aguycalled:
This PR adds support for a new config parameter:

- banversion

Example of use:

banversion=/NavCoin:4.1.0/
banversion=/NavCoin:4.1.1/

Nodes broadcasting the specified versions will be banned for 24h. This helps to reduce an excessive traffic in the nodes caused by old wallets constantly sending blocks with old consensus rules.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -6081,6 +6081,25 @@ bool static ProcessMessage(CNode* pfrom, string strCommand, CDataStream& vRecv, 
                   pfrom->cleanSubVer, pfrom->nVersion,
                   pfrom->nStartingHeight, addrMe.ToString(), pfrom->id,
                   remoteAddr);
+	    
+        std::vector<std::string> vBannedVersions = mapMultiArgs["-banversion"];
+        bool fBanned = false;
+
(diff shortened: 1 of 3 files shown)


src/main.cpp
```diff
@@ -6081,6 +6081,25 @@ bool static ProcessMessage(CNode* pfrom, string strCommand, CDataStream& vRecv,
                   pfrom->cleanSubVer, pfrom->nVersion,
                   pfrom->nStartingHeight, addrMe.ToString(), pfrom->id,
                   remoteAddr);
+	    
+        std::vector<std::string> vBannedVersions = mapMultiArgs["-banversion"];
+        bool fBanned = false;
+
+        for (unsigned int i = 0; i <= vBannedVersions.size(); i++)
```
matt-auckland: I think we should ban like this instead:
`if nodeVer <= bannedVer`
Not sure how we're representing versions in the codebase 
If it's like this `04002001` it should be easy to make this change)

matt-auckland:
lol didn't notice this branch was already merged

PR #242, opened 2018-07-13
validateaddress doc typo for isstakable attribute

dantudor:
The help for validateaddress specified the ismine field twice. The second should be isstakable


red010b37:
Nice!

PR #262, opened 2018-08-03
Secp256k1 4.3.0 update

red010b37:
Updates lib and build and configure


configure.ac
```diff
@@ -21,6 +21,19 @@ NAVCOIN_TX_NAME=navcoin-tx
 
 AC_CANONICAL_HOST
 
+AC_MSG_NOTICE([------------------------------------])
+AC_MSG_NOTICE([------------------------------------])
+AC_MSG_NOTICE([               __                   ])
+AC_MSG_NOTICE([     |\ | _   /   _ . _             ])
+AC_MSG_NOTICE([     | \|(_|\/\__(_)|| |            ])
+AC_MSG_NOTICE([                                    ])
+AC_MSG_NOTICE([     Setting up build for $host     ])
```
aguycalled: the line looks out when the host too long, i'd suggest breaking it in two lines

aguycalled:
everything looks good. could you please include notes on /doc/release-notes/release-notes-4.3.0.md about the reason why you updated, which is the new version and where did you get it from (my impression is from latest bitcoin)
could you also post comments about your testing on the binaries of the different architectures and gitian builds. 
then we are ready to merge ;) good job

red010b37:
I will fix the line update the release notes tomorrow for you. :)

red010b37:
@aguycalled I have updated the release notes, I also updated the OSX build notes and added a sh script that helps non-osx users get the tar ready to be used in the gitian builder system,

PR #276, opened 2018-08-14
WIP: Use navexplorer.com as the explorer to view transactions

dantudor:
Updated nav core wallet to use navexplorer.com as the block explorer for user transactions in the qt GUI.


craigmacgregor:
looks good to me, will build and test

PR #301, opened 2018-09-28
Allow getrawtransaction verbose mode to be activated with int or bool

dantudor:
NBitcoin compares verbose mode using boolean. 
Extending getrawtranasction to use both int and boolean flags.

### DIFF
--- a/src/rpc/rawtransaction.cpp
+++ b/src/rpc/rawtransaction.cpp
@@ -319,7 +319,7 @@ UniValue getrawtransaction(const UniValue& params, bool fHelp)
     bool fVerbose = false;
     if (params.size() > 1 && !params[1].isNull()) {
         if (params[1].isNum()) {
-            if (params[1].get_int() == 1) {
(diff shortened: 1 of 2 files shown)


src/rpc/rawtransaction.cpp
```diff
@@ -310,13 +310,25 @@ UniValue getrawtransaction(const UniValue& params, bool fHelp)
...
-    if (params.size() > 1)
-        fVerbose = (params[1].get_int() != 0);
+    if (params.size() > 1 && !params[1].isNull()) {
+        if (params[1].isNum() && params[1].get_int() == 1) {
+            fVerbose = true;
+        }
+        else if(params[1].isBool() && params[1].isTrue()) {
+            fVerbose = true;
+        }
+        else {
```
aguycalled: what would happen here if params[1] is 0 or false?

dantudor: Yep. Only raise the error if the type is not int or bool and not dependent on value

PR #333, opened 2018-12-04
New quorum for the Community Fund

aguycalled:
This Pull Request proposes a Fork signalled by version bit 17 introducing a reduction of the required quorum for the Community Fund in the second half of the votings.

Needs to be tested:
- Soft fork actives correctly, old blocks are rejected.
- Proposals and Payment Requests have version 3 with the Fork active, version 2 otherwise.
- Proposals and Payment Requests with wrong version are rejected.
- Proposals and Payment Requests with version 2 use the legacy quorum.
- Proposals and Payment Requests with version 3 use the new quorum system: when the voting cycle is > (total number of periods / 2) then use reduced quorum 40%, otherwise normal quorum 50%


dantudor:
**Current model:**
At least 50% of the blockcycle needs to vote (quorum)
For a proposal to be accepted:
 - At least 70% of those votes must be `yes` to accept the proposal
 - At least 70% of those votes must be `no` to reject the vote

After 5 cycles if there has been no decision then the proposal is rejected

**Example Scenario**
A proposal has been created and has met the voting quorum with total votes at 60%
70% of those votes are `yes`, 30% of those voters are `no`. The proposal will be accepted.

The 30% `no` voters do not want the proposal to complete and petition Discord and Reddit to vote `no` on the proposal.
More people vote `no, but the proposal still passes.

To actually get the proposal to not be accepted, the no voters instead need to petition Reddit and Discord to remove their votes.
They change their `no` vote to `abstain`, the quorum reduces to less than 50% and the proposal is not accepted.

This to me does not seem like the intended scenario.




**Suggested model:**
At least 35% of the blockcycle must vote `yes` (quorum)
Or at least 35% of the blockcycle must vote `no` (quorum)

For a proposal to be accepted:
 - `Yes` votes must out weigh `no` votes by 2-1
 - `No` votes must out weigh `yes` votes by 2-1

After 5 cycles if there has been no decision then the proposal is rejected

**Scenario**
A proposal has been created and has met the voting quorum with total votes at 35% `yes` That translates to 7056 `yes` votes.
There are 2800 `no` votes. The ratio of yes:no votes is greater than 2:1 and therefore the proposal will be `accepted`

The 2800 `no` voters do not want the proposal to complete and petition Discord and Reddit to vote `no` on the proposal.
The `no` voters need to get 729 new no votes to force the proposal to not be accepted and run into the next cycle.

If the no voters close the gap and get the 729 votes taking their total to 3529. 
The ratio is now below 2:1 and the proposal will not be accepted.
For every new `no` vote from now, the `yes` voters need to gain 2 new votes to maintain the 2-1 ratio and accept the proposal.


**Applying the second model to a selection of existing proposals:**

1. Navcoin sticker creation for the community
40% of the total voters have voted `yes`
The ratio of `yes`, `no` votes is greater than 2-1
The proposal will pass

2. Cryptocandor youtube review
36% of the total voters have votes `yes`
The ratio of `yes`, `no` votes is greater than 2-1

3. Navcoin next wallet (completed)
20% of the total voters have voted `yes`
4% of the total voters have voted `no`
The proposal has not met the quorum and will not be accepted.
The `no` voters are not contributing to the quorum for acceptance.
The `no` voters can confidently continue to vote no and not help pass the proposal.


There was suggestion that a `no` vote isn't a `no` vote it's an `expire this proposal early`. 
I'd disagree with the statement. A no vote is a staker showing their disapproval for a proposal.
We should encourage as many stakers as possible to participate in the community fund.
Encouraging `no` voters to not have a voice and to abstain to reject a proposal does not send the correct message to the proposal owner.

**A `yes` vote states:** 
"We the community value your proposal and would like the proposal to be accepted"

**A `no` vote states:** 
"We the community do not value your proposal and would like the proposal to be rejected"

**An `abstain` vote states:** 
"We the community do not hold strong feeling for this proposal. We need persuading either way"

sakdeniz:
I also agree that the 50% participation rate is currently high. At the moment, the participation rate is around 40%. Therefore, it would be appropriate to reduce the participation rate to 40%.

PR #335, opened 2018-12-05
Header spam protection

aguycalled:


craigmacgregor:
Docker tests are showing this fix is effective at blocking headers spam.

I've also tested the added flags and can confirm that they are working as expected.

PR #348, opened 2018-12-10
Fixed coldstaking_spending.py 

djksa:
The test outcome was previously non-deterministic due to changing minimum fees for the same type of transaction, the test used to also fail in general due to the send_raw_transaction() function definition.

I have added more transaction padding considering this test is not testing the amount we send is correct but is testing whether we can send and whether funds have been received. I have made some assert statements more strict as well.

### DIFF
--- a/qa/rpc-tests/coldstaking_spending.py
+++ b/qa/rpc-tests/coldstaking_spending.py
@@ -3,11 +3,10 @@
 # distributed under the mit software license, see the accompanying
 # file copying or http://www.opensource.org/licenses/mit-license.php.
 import decimal
-import time
 from test_framework.test_framework import NavCoinTestFramework
 from test_framework.util import *
-
-class SendingFromColdStaking(NavCoinTestFramework):
+#/TODO leave comments for how much is sent in each func call
+class ColdStakingSpending(NavCoinTestFramework):
     """Tests spending and staking to/from a spending wallet."""
 
     # set up num of nodes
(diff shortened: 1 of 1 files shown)


qa/rpc-tests/coldstaking_spending.py
```diff
@@ -3,11 +3,10 @@
 # distributed under the mit software license, see the accompanying
 # file copying or http://www.opensource.org/licenses/mit-license.php.
 import decimal
-import time
 from test_framework.test_framework import NavCoinTestFramework
 from test_framework.util import *
-
-class SendingFromColdStaking(NavCoinTestFramework):
+#/TODO leave comments for how much is sent in each func call
```
matt-auckland: Is this done? If so we should remove this line

```suggestion
```

qa/rpc-tests/coldstaking_spending.py
```diff
@@ -154,6 +148,8 @@ def run_test(self):
         assert(send_worked == True)
         
         slow_gen(self.nodes[0], 1)
+        # self.sync_all()
```
matt-auckland: If we aren't using this code we should remove it
```suggestion
```

qa/rpc-tests/coldstaking_spending.py
```diff
@@ -154,6 +148,8 @@ def run_test(self):
...
+
```
matt-auckland:
```suggestion
```

qa/rpc-tests/coldstaking_spending.py
```diff
@@ -3,13 +3,11 @@
...
 import decimal
-import time
 from test_framework.test_framework import NavCoinTestFramework
 from test_framework.util import *
-
-class SendingFromColdStaking(NavCoinTestFramework):
+class ColdStakingSpending(NavCoinTestFramework):
     """Tests spending and staking to/from a spending wallet."""
-
+
```
matt-auckland:
```suggestion
```
One last pedantic change

PR #355, opened 2018-12-11
Cfund new propoal alert toolbar [X] button fix

matt-auckland:
The CFund new proposal alert pop currently cannot be closed using the [X] button, this patch attempts to remedy that.

QT seems to have a bug wherein if there is no button with `QMessageBox::RejectRole` assigned to it, the [X] button in the toolbar will not work.

To fix this I have added a close button with this functionality and hidden it, which allows the [X] to work as intended but also prevents people from dismissing the window instantly by clicking on a big 'Close' button.

### DIFF
--- a/src/NavCoin4-Qt.files
+++ b/src/NavCoin4-Qt.files
@@ -91,6 +91,9 @@ protocol.cpp
 protocol.h
 pubkey.cpp
 pubkey.h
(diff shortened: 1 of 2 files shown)


craigmacgregor:
approved pending travis build.

PR #367, opened 2018-12-14
Change ifelse branches in CountVotes()

aguycalled:
This PR fixes a bug in the logic of the CountVotes() function which can cause a proposal to morph from a Expired state back to an Accepted state when a node reindex the chainstate db and the proposal expired being accepted.

The fix involves changing the structure of some conditionals, so in the case when a proposal state and status do not match the next else if branches are not triggered. It also considers a proposal unexpired state as always expired. See the changes list for more details.

This issue is critical as it can cause two different states in different nodes and a fork in the network.

### DIFF
--- a/qa/rpc-tests/cfund-proposal-state-accept.py
+++ b/qa/rpc-tests/cfund-proposal-state-accept.py
@@ -23,7 +23,9 @@ class CommunityFundProposalStateTest(NavCoinTestFramework):
     def run_test(self):
         activate_cfund(self.nodes[0])
 
-        proposalid0 = self.nodes[0].createproposal(self.nodes[0].getnewaddress(), 1, 3600, "test")["hash"]
+        proposal_duration = 3
+
+        proposalid0 = self.nodes[0].createproposal(self.nodes[0].getnewaddress(), 1, proposal_duration, "test")["hash"]
         slow_gen(self.nodes[0], 1)
 
         start_new_cycle(self.nodes[0])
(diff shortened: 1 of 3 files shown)


craigmacgregor:
@marcus290 what have you tested on this fix? do we need to write or update any unit tests?

craigmacgregor:
I've compiled and run the tests and they all pass. I also compiled the master branch and ran the tests against that and they pass there too. So I'm unable to say just by the tests that the bug is fixed. I am wondering if when we find a bug like this, the first step should be to write a test that fails when the bug is present but should pass when the bug is fixed. Therefore we can definitively say that the bug is fixed. I am going to see if i can modify the test to reproduce the issue on master and resolve it on the patch branch.

marcus290:
@craigmacgregor there are two issues - 
1) Reindexing will recalculate expired proposals.
2) Expired proposals will change to expired (`fState: 3` under listproposals) but the `status` will show "accepted waiting for end of voting period". At the end of the voting cycle, the proposal will be reaccepted. 


Steps to reproduce the issues:
- open testnet and let the wallet fully sync. Then reopen the wallet with -reindex-chainstate. A previously expired proposal will reappear when running `listproposals` because its expiry has been recalculated.
- create a short life proposal on regnet and let it expire. Then check the status upon expiry and after the next voting cycle.

I manually retested both these issues with this commit and they are fixed according to my tests.

Agreed we should write tests.

PR #369, opened 2018-12-14
Fix fixed stake reward check

aguycalled:
The stake reward, even if fixed, includes the fees of the transactions included in the block.

The removed check does not take the fees making the network reject all the blocks which include a transaction. At this point of the function, the total amount of fees is still unknown.

This is safely removed as the check is properly done in line 3322.


craigmacgregor:
@aguycalled do we want to write a test that tries to construct a block which has an incorrect stake amount and test whether it is accepted on the network, or are we confident this is covered correctly by line 3322?

aguycalled:
Line 3322 covers correctly the check, but we need general tests for our whole POS implementation, there's none currently.

PR #370, opened 2018-12-14
Fixes Payment Request payout when Proposal is expired

aguycalled:
Adds a new state for proposals (PENDING_VOTING_PREQ). This transitory state
will be asigned to expired proposals with payment requests pending being
voted. Once a expired proposal runs out of pending payment requests,
the proposal will move to EXPIRED state.
Payment Requests voting can only start while the state of the parent Proposal
is ACCEPTED.
Payouts to Payment Requests are only authorized when a Proposal is in
ACCEPTED or PENDING_VOTING_PREQ state.
Includes RPC test.


matt-auckland:
I'm checking this out and testing it

PR #371, opened 2018-12-14
v4.5.1

aguycalled:
- Bumps version to v4.5.1
- Sets a fork height at 2722100
- Requires blocks to signal version bit 21 when fork happens.
Merge after #367 #369 and #370 or when its related issues are fixed.


src/versionbits.h
```diff
@@ -59,6 +60,7 @@ static const int32_t nCFundAccSpreadVersionMask = 0x00004000;
 static const int32_t nCFundAmountV2Mask = 0x00010000;
 static const int32_t nCFundReducedQuorumMask = 0x00020000;
 static const int32_t nStaticRewardVersionMask = 0x00008000;
+static const int32_t nV451ForkMask = 0x00100000;
```
craigmacgregor: 0x00100000 = 0000 0000 0000 0000 0001 0000 0000 0000 0000 0000

which is version bit 20 if i counted right, the comment above says version bit 21. Not sure if it matters?

aguycalled: you are right

craigmacgregor: so i guess we just need to update the comment? everything else works off the bit mask anyway right?

aguycalled: yes

craigmacgregor:
I just wrote a test that checks the version bit activates after the specified height. Is there anything else we want to unit test? I can try to setup a test with two nodes where one tries to broadcast an obsolete block version, but I haven't used the test framework to broadcast manipulated blocks. Might take some time. Not sure if it's worth doing. Has the obsolete rejection been manually tested?

aguycalled:
it hasn't been manually tested yet, testing the other pr currently

craigmacgregor:
Okay, ill see what i can do to test this in the morning either by unit test or manually

matt-auckland:
I am manually testing this

matt-auckland:
Fork tested, appears to be working.

## Test method

Set up two nodes: 1x 4.5.1, 1x 4.5.0
Generated blocks till the fork, they both remained in sync.
After the fork blocks from the 4.5.0 node are not accepted by the 4.5.1 node as expected.
The 4.5.0 node continues to accept blocks from the 4.5.1 node and will reorg and discard it's own blocks.

src/main.cpp
```diff
@@ -4954,9 +4960,13 @@ bool ContextualCheckBlockHeader(const CBlockHeader& block, CValidationState& sta
                         "rejected no static reward block");
 
    if((block.nVersion & nCFundReducedQuorumMask) != nCFundReducedQuorumMask && IsReducedCFundQuorumEnabled(pindexPrev,Params().GetConsensus()))
-     return state.Invalid(false, REJECT_OBSOLETE, strprintf("bad-version(0x%08x)", block.nVersion),
+       return state.Invalid(false, REJECT_OBSOLETE, strprintf("bad-version(0x%08x)", block.nVersion),
                         "rejected no reduced quorum block");
 
+   if((block.nVersion & nV451ForkMask) != nV451ForkMask && pindexPrev->nHeight >= Params().GetConsensus().nHeightv451Fork)
+       return state.Invalid(false, REJECT_OBSOLETE, strprintf("bad-version(0x%08x)", block.nVersion),
+                        "rejected no v4.5.1 block");
```
matt-auckland:
```suggestion
                        "rejected, block version isn't v4.5.1");
```

matt-auckland: This seems to be the rejection message for when we get a block, if so, this wording might be better

PR #372, opened 2018-12-15
Test which reproduces the expired morph bug - DONT MERGE

craigmacgregor:
This test reproduces the expired morph bug and correctly fails on 4.5.0


matt-auckland:
Can we close this?

PR #380, opened 2018-12-17
RPC paymentrequestvotelist and proposalvotelist change

aguycalled:
- Change output to show JSON instead of string


matt-auckland:
Looks good, will approve after testing

PR #390, opened 2019-01-05
fixes #388: cache cfund votes and add bench log

### LINKED ISSUE #388: Wallet loading blocks slowly since Community Fund update (~ 3 blocks per/sec).
Wallet loading blocks slowly (~ 3 blocks per/sec).

Wallet loading blocks slowly since the Community Fund update.

NavCoin NEXT with 4.5.1 build

Windows 10, intel i5 CPU, 16 GB Ram

aguycalled:
Solves #388.

Keeps a cache of the community fund votes to avoid an unnecessary recalculation per block which was causing extreme slow synchronization in the order of 0,25-0,3s per block.

After applying the patch the CountVotes() function takes an average of 32ms.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -3871,7 +3871,7 @@ void CountVotes(CValidationState& state, CBlockIndex *pindexNew, bool fUndo)
 
     std::map<uint256, bool> vSeen;
 
-    if (fUndo || nBlocks == 1) {
+    if (fUndo || nBlocks == 1 || vCacheProposalsToUpdate.empty() || vCachePaymentRequestToUpdate.empty()) {
         vCacheProposalsToUpdate.clear();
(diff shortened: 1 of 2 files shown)


marcus290:
@aguycalled @matt-auckland I'm getting a mismatch between the no. of votes shown with `listproposals` and `cfundstats`, where `listproposals` is showing much fewer votes. Do you get the same thing?

The sync is much faster indeed.

aguycalled:
good catch. committed f1154b4

craigmacgregor:
The sync speed is good. I'm getting through a few hundred blocks per second or more. 

@marcus290 I'll leave it to you to confirm if you're happy with the fix for the votes stats.

aguycalled:
can we merge this?

PR #392, opened 2019-01-07
Added rpc help test

hdnsimpson:
Created new rpc test: rpc-help
Tests the ability to generate help at runtime as well as all expected help commands are available


matt-auckland:
Can we have another look at this? Travis seems to crash when it reaches this test

PR #393, opened 2019-01-07
v452

craigmacgregor:
Updated versioning to 4.5.2

Enforce version bit 21 for all blocks after 2800000 (approximately Friday 18 January 2019).

The RPC test has also been made more generic so you just pass in the version bit and activation height rather than needing a function for each version bit.

We will also want to add functionality which checks if height has been reached when rejecting blocks with invalid CFund payment requests and any other fixes deployed in v452.


aguycalled:
this pr is missing the patch from https://github.com/aguycalled/navcoin-core/tree/fix-preq-payout which should be conditioned to nHeight >= Params().GetConsensus().nHeightv452Fork

Issue #391, opened 2019-01-05
Community fund proposal votingCycle does not stop incrementing when the state is terminal


dantudor:
When a proposal state has changed to a terminal state the votingCycle should remain at the previous value and no longer increment for future cycles.

Currently when `navcoin-cli getproposal bdc588af828557deb114f67c5dfa87e44e4f25b142ac091aa6d65ee653cbf44a` is executed, a votingCycle value of 7.

This value should have remained at the votingCycle value of acceptance.

marcus290:
by design, the votingCycle goes +1 over. `IsExpired()` relies on this to check for voting period expiry.

Otherwise, the votingCycle value shouldn't go any higher. I wasn't able to replicate this in my own 4.5.1 wallet. `getproposal bdc588af828557deb114f67c5dfa87e44e4f25b142ac091aa6d65ee653cbf44a` gave me a voting cycle of 5 not 7. Any other way to replicate this?

Nevertheless, I think we should stop counting those votes in votingCycle 7.

aguycalled:
Yes your analysis is correct. With the current implementation the proposal can get votes in `nCyclesPaymentRequestVoting+1` but will never morph to an state different than `EXPIRED` so the votes have no effect.

Ideally the proposal should morph in a `EXPIRED WAITING FOR END OF VOTING PERIOD` state when on cycle `nCyclesPaymentRequestVoting` and morph to `EXPIRED` on `nCyclesPaymentRequestVoting+1` if it was not `ACCEPTED` or `REJECTED` on cycle `nCyclesPaymentRequestVoting`.

The fix would be to change the check in IsExpired from:

```nVotingCycle > Params().GetConsensus().nCyclesProposalVoting ```

to:

```nVotingCycle >= Params().GetConsensus().nCyclesProposalVoting ```

And to break the if-elseif branch in CountVotes() to

```
if(proposal.IsExpired(pindexNew->GetBlockTime())){
        ....
}
if(proposal.IsRejected()){
        ....
}
...
```

Example of code for my proposal of fix (missing the fork part) at https://github.com/aguycalled/navcoin-core/tree/cfund-last-cycle

marcus290:
Before proposals/prequests can be accepted, proposals will check `(proposal.fState == CFund::NIL || proposal.fState == CFund::PENDING_FUNDS)` and payment requests will check `(prequest.fState == CFund::NIL)` .

[...]

PR #396, opened 2019-01-09
Interim fix for #391 cfund votingCycle increments

marcus290:
This PR is an interim fix for #391.

Uses https://github.com/aguycalled/navcoin-core/commit/9c1db6473f7457e09e893625e40968c728930c18 to stop votingCycles being recalculated if there is a reorg. Adds a test to check this, which fails in the current master branch but passes with this patch.

Expired proposals and payment requests will still increment +1 over their maximum number of cycles as per the existing cfund design.

Also stops votes from being counted in the votingCycle `nCyclesPaymentRequestVoting + 1` and `nCyclesProposalVoting + 1`.


aguycalled:
ideally tests should be duplicated, with one of them launching the node with flag rejectversionbit=n to verify behaviour in pre-fork scenarios and the other verifying behaviour in post-fork scenarios

marcus290:
I modifed proposal.toJson() and prequest.toJson() to show min of `nVotingCycle` and `

craigmacgregor:
Isn't this change cosmetic only? Do we need to attach the change to the version bit change?

src/consensus/cfund.cpp
```diff
@@ -384,18 +384,47 @@ bool CFund::CProposal::IsRejected() const {
...
 }
 
+bool CFund::CProposal::CanVote() const {
+    return (fState == NIL) && (nVotingCycle <= Params().GetConsensus().nCyclesProposalVoting);
+}
+
 bool CFund::CProposal::IsExpired(uint32_t currentTime) const {
     if(nVersion >= 2) {
         if (fState == ACCEPTED && mapBlockIndex.count(blockhash) > 0) {
             CBlockIndex* pblockindex = mapBlockIndex[blockhash];
```
matt-auckland:
```suggestion
            CBlockIndex* pBlockIndex = mapBlockIndex[blockhash];
```
🐪case?

src/consensus/cfund.cpp
```diff
@@ -384,18 +384,47 @@ bool CFund::CProposal::IsRejected() const {
...
             return (pblockindex->GetBlockTime() + nDeadline < currentTime);
         }
-        return (fState == EXPIRED) || (fState == PENDING_VOTING_PREQ) || (nVotingCycle > Params().GetConsensus().nCyclesProposalVoting && (CanVote() || fState == EXPIRED));
+        return (fState == EXPIRED) || (fState == PENDING_VOTING_PREQ) || (nVotingCycle > Params().GetConsensus().nCyclesProposalVoting && (fState == NIL || fState == EXPIRED));
```
matt-auckland:
```suggestion
        return (fState == EXPIRED) || (fState == PENDING_VOTING_PREQ) || (nVotingCycle > Params().GetConsensus().nCyclesProposalVoting && fState == NIL);
```
If `fState == EXPIRED` is true then this boolean statement will always return before it gets to this check at the end because this check exists at the start

src/consensus/cfund.cpp
```diff
@@ -384,18 +384,47 @@ bool CFund::CProposal::IsRejected() const {
...
+            sFlags += " waiting for end of voting period";
+    }
+    if(IsRejected()) {
+        sFlags = "rejected";
+        if(fState != REJECTED)
+            sFlags += " waiting for end of voting period";
+    }
+    if(currentTime > 0 && IsExpired(currentTime)) {
+        sFlags = "expired";
+        if(fState != EXPIRED && nVotingCycle <= Params().GetConsensus().nCyclesProposalVoting)
```
matt-auckland: We use this check `nVotingCycle <= Params().GetConsensus().nCyclesProposalVoting` all over the place, but it's not really clear as to what it's purpose is, if a proposal is `EXPIRED` why do we need to check the voting cycle is? I understand it's because we don't update the `fState` until the end of the 7th cycle, however to make this clearer I think we should do one of two things:
- Add comments in the code around these checks so it's clear what their purpose is (easiest method but kind of annoying, not keen on this)
- Refactor this check out into a function that has a descriptive name and potentially a single comment in the function that describes it's purpose. 
|e.g. ```
checkIfVotingCycleCapExceeded() {
// this function checks if a proposal who's fState has not yet been set to expired ....etc
}
```

File: src/consensus/cfund.cpp [hunk 4]
```diff
@@ -416,7 +445,7 @@ void CFund::CProposal::ToJson(UniValue& ret) const {
     }
     ret.push_back(Pair("votesYes", nVotesYes));
     ret.push_back(Pair("votesNo", nVotesNo));
-    ret.push_back(Pair("votingCycle", (uint64_t)nVotingCycle));
+    ret.push_back(Pair("votingCycle", (uint64_t)std::min(nVotingCycle, Params().GetConsensus().nCyclesProposalVoting)));
```
matt-auckland:
```suggestion
    ret.push_back(Pair("votingCycle", (uint64_t)std::min(nVotingCycle, Params().GetConsensus().nCyclesProposalVoting))); // We return the smaller of the current voting cycle or the max voting cycle possible so we don't return the mysterious 7th voting cycle
```

matt-auckland: We should have a descriptive comment in this vein for this line of code and the similar one for payment requests

src/consensus/cfund.cpp
```diff
@@ -443,7 +472,7 @@ void CFund::CPaymentRequest::ToJson(UniValue& ret) const {
     ret.push_back(Pair("requestedAmount", FormatMoney(nAmount)));
     ret.push_back(Pair("votesYes", nVotesYes));
     ret.push_back(Pair("votesNo", nVotesNo));
-    ret.push_back(Pair("votingCycle", (uint64_t)nVotingCycle));
+    ret.push_back(Pair("votingCycle", (uint64_t)std::min(nVotingCycle, Params().GetConsensus().nCyclesPaymentRequestVoting)));
```
matt-auckland: I think we should add a comment to the code as suggested above

matt-auckland:
I've reviewed the code except for the tests, the code looks good for the most part. 
Will check the tests later. For now see the comments I've left.

PR #400, opened 2019-01-20
Mnemonic support

aguycalled:
This PR adds a new RPC command to export the existing master private key encoded as a mnemonic:

dumpmnemonic

It admits an argument specifying the language.

Support for two new wallet options (-importmnemonic and -mnemoniclanguage) have also been added to allow to create a new wallet from the specified mnemonic.

Testing needed.


matt-auckland:
cACK, will test this when I have spare time

src/mnemonic/mnemonic.cpp
```diff
@@ -0,0 +1,264 @@
...
+        {
+            words.push_back(sentence.substr(prevPos, i - prevPos));
+            prevPos = i + 1;
+        }
+    }
+    return words;
+}
+
+
+//long_hash decode_mnemonic(const word_list& mnemonic,
```
matt-auckland: Is there a reason we need to leave this code in?

src/qt/navcoingui.cpp
```diff
@@ -150,13 +150,13 @@ NavCoinGUI::NavCoinGUI(const PlatformStyle *platformStyle, const NetworkStyle *n
     prevBlocks(0),
     spinnerFrame(0),
     fDontShowAgain(false),
-    lastDialogShown(0),
     unlockWalletAction(0),
     lockWalletAction(0),
     toggleStakingAction(0),
```
matt-auckland: What is this code for?

src/qt/navtechinit.cpp
```diff
@@ -25,7 +25,6 @@ NavTechInit::~NavTechInit()
 void NavTechInit::ShowNavtechIntro(bool exitAfter)
 {
     NavTechInit navtechinit;
```
matt-auckland: What is this code for?

craigmacgregor:
@matt-auckland going to evaluate functionality and report back

marcus290:
I manually tested `dumpmnemonic` in different languages and `-importmnemonic` and `-mnemoniclanguage` and it works for me.

Just wondering, why not add support for the remaining languages that are already in dictionary.cpp?

I also think 80512622b488ee2d232b797ace130fea6b05d15e maybe should be moved out to a separate commit.

Issue #399, opened 2019-01-17
Option to be able to stake to a customised address


Namugikamawara:
Hello, 

I want to define a  customised address on which the stakes will be automatic sent;
Like this,i I have one wallet to stake ,other to spend.
Thanks.

chasingkirkjufell:
Isn't that just cold staking? Set one wallet as your staking address and another as your spending address. All the stakes are only spendable with spending address wallet.

aguycalled:
He means without cold staking, to automatically send only the rewards to a different address.

PR #401, opened 2019-01-20
Accumulation of staking rewards in different address

aguycalled:
In response to #399 , this PR introduces support for the -stakingaddress argument which sets a navcoin address where the staking rewards are accumulated.

Not compatible with cold staking.


marcus290:
@aguycalled function works for me but the wallet gui shows the wrong destination address for me:

screenshot from 2019-01-22 13-31-49

Also, seems like gettransaction rpc command will show the wrong address (I think the `vin` address):
screenshot from 2019-01-22 13-33-45

I think `stop = true; // only one coinstake output` in rpcwallet.cpp (line 1908) will cause some issues.

src/compat/byteswap.h
```diff
@@ -1,34 +1,47 @@
-// Copyright (c) 2014 The Bitcoin developers
+// Copyright (c) 2014-2018 The Bitcoin Core developers
```
matt-auckland: Is this related to staking? Seems like it's been mixed in

src/wallet/wallet.cpp
```diff
@@ -546,6 +546,14 @@ bool CWallet::CreateCoinStake(const CKeyStore& keystore, unsigned int nBits, int
         txNew.vout[1].nValue = blockValue;
     }
 
+    if (GetArg("-stakingaddress", "") != "" && !txNew.vout[txNew.vout.size()-1].scriptPubKey.IsColdStaking()) {
```
matt-auckland: When we validate a block do we check anything specific about a staking output?

craigmacgregor:
This is a great MVP but i would argue that we should add a little more functionality before releasing it into the main net.

As a user I can have multiple staking addresses which i might want to send their rewards to different reward addresses. I think we should make this value more of a map rather than a single value where i could specify "staking address:reward address" pairs. Or have an "all:reward address" which handles every staking address not otherwise specified in its own pair.

matt-auckland:
Tested and it appears to work. I'm not sure what the syntax is for staking to multiple addresses, this could do with documentation.

PR #386, opened 2018-12-21
Update rpc-tests (Don't merge)

djksa:
▶️ [NEW PULL REQUEST HERE](https://github.com/NAVCoin/navcoin-core/pull/405) ◀️ 

Fixed certain rpc-tests from Issue #358

Issue #358, opened 2018-12-12
Fix RPC tests


matt-auckland:
We currently have a number of RPC tests inherited from BTC, however they are in various states of 'working'. It would be good if we could get them all in a functional state.

Some of the tests will probably succeed without any need for adjustment and others will fail. Regardless we will need to look over the tests to check that they are still testing the correct things (for example we may have adjusted something so that it differs from bitcoin but the test still passes)

For reference the tests are all located in `qa/rpc-tests/` and are called:
```
p2p-fullblocktest.py,
walletbackup.py,
bip68-112-113-p2p.py,
wallet.py,
wallet-hd.py,
listtransactions.py,
receivedby.py,
mempool_resurrect_test.py,
txn_doublespend.py --mineblock,
txn_clone.py,
getchaintips.py,
rawtransactions.py,
rest.py,
mempool_spendcoinbase.py,
mempool_reorg.py,
mempool_limit.py,
httpbasics.py,
multi_rpc.py,
zapwallettxes.py,
proxy_test.py,
merkle_blocks.py,
fundrawtransaction.py,
signrawtransactions.py,
nodehandling.py,
reindex.py,
addressindex.py,
timestampindex.py,
spentindex.py,
txindex.py,
decodescript.py,
blockchain.py,
disablewallet.py,
sendheaders.py,
keypool.py,
prioritise_transaction.py,
invalidblockrequest.py,
invalidtxrequest.py,
abandonconflict.py,
p2p-versionbits-warning.py,
p2p-segwit.py,
segwit.py,
importprunedfunds.py,
signmessages.py
```

djksa:
Investigating this.

hdnsimpson:
Currently fixed:
getchaintips.py
rest.py

dy5es41:
fixed:
- wallet-hd.py
tests that seem to work with no changes (pass true)
- disablewallet.py
- nodehandling.py
notes on other tests:
- after editing walletbackup.py it seems to timeout due to the path to wallet.dat being incorrect, which is a general theme amongst several tests. changing the path from `/node1/regnet/wallet.dat` to `/node1/devnet/wallet.dat` seems to fix timeout errors

mxaddict:
We've opened a bounty issue for this here: #519

PR #405, opened 2019-01-29
Fixed rpc tests

djksa:
Fixed certain rpc-tests from Issue #358.

This pull-request contains a branch based off of the branch `update-rpc-tests` from my original pull request [here](https://github.com/NAVCoin/navcoin-core/pull/386). We decided that it would be safer and easier to cherry-pick the working rpc-tests from the `update-rpc-tests` branch and to manually copy them over to the `fixed-rpc-tests` branch rather than attempt to resolve the old pull-request.


marcus290:
To summarise, all uncommented tests in `rpc-tests.py` are newly implemented into the test suite.

It seems that using `initialize_chain()` instead of `initialize_chain_clean()` in `util.py` was causing TimeoutErrors in some tests, which is addressed in 5cb057e.

PR #409, opened 2019-01-30
disable darkmode on Mojave

### LINKED ISSUE #403: Issue with MacOS Mojave dark mode
The GUI does not render correctly the colors when using MacOS Mojave in dark mode.

Examples:

image

Address not showing, background for text field amount is black

image

Black background

marcus290:
This patch fixes #403 by disabling dark mode in MacOS Mojave. @matt-auckland found the same issue at https://github.com/bitcoin/bitcoin/pull/14593 where the conclusion was to disable dark mode until Qt has further support for darkmode. The fix is taken from there.

### DIFF
--- a/share/qt/Info.plist.in
+++ b/share/qt/Info.plist.in
@@ -100,6 +100,9 @@
   <key>LSAppNapIsDisabled</key>
     <string>True</string>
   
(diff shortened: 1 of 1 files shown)


aguycalled:
tested, works

PR #412, opened 2019-02-03
ZeroCT - Full implementation with Transaction Amount obfuscation support and ZeroCT Staking

skreener:
- Adds support for Zerocoin Mint and Spend operations
- New type of address to anonymously receive coins
- Deterministic private address generated from master private key of the wallet
- Use of GMP for Bignum operations
- Use of threading for Mint and Spend verification
- Witnesser thread for pre-computing of coin witnesses
- Monitoring of public and private coin supply
- Staking with private coins
- Calculation of staking weight based on private coins
- Fees collected from public transactions are accumulated and only earned by public coin stakes
- Stakes using coins from the private pool earn in exclusive accumulated fees from private transactions
- New default data folder: NavCoin5 / navcoin5
- Testnet uses testnet4 to store chain data.
- Old NavTech functionality and references are removed
- GUI-level integrations of private transactions functionality
- Encryption of the obfuscation parameters for encrypted wallets
- New rpc commands:
  - `getprivateaddress`
  - `privatesendtoaddress`
  - `dumpprivateparameters`
- New daemon arguments:
  - `-enablewitnesser`: enables witnesser thread to precompute coin witnesses (default true)
  - `-witnesser_block_snapshot`: count of blocks to snapshot and backup witnesses value
  - `-witnesser_blocks_per_round`: count of checked blocks per witnessing round
- Includes libzerocoin test units

# Consensus rules
- Transactions do not admit combinations of public and private coins as inputs
- Serial numbers and Mint coin values can only be spent once
- Spend/mint fee: per transaction weight

# To do
- Include obfuscation parameters in `dumpwallet` output
- When a transaction size is too big, split it in different transactions containing the different spend proofs
- Accommodate wallet-related RPC commands to the new private transactions functionality
- Redesign/improve UX in the GUI
- Remove critical log entries which could compromise user's privacy
- Improve proof's verification time/general performance
- Testing ;-)


mxaddict:
@aguycalled should we close this PR since this work was cancelled?

PR #428, opened 2019-02-27
Core Wallet Community Fund UI Improvement

hdnsimpson:
image

 - Added a Community Fund tab to the core wallet
 - Ability to view and filter proposals and payment request
 - Proposals and payment requests can be voted on and created
 - Removed the old Community Fund UI


src/qt/communityfundsuccessdialog.cpp
```diff
@@ -0,0 +1,45 @@
...
+    prequest(0)
+{
+    ui->setupUi(this);
+    connect(ui->pushButtonClose, SIGNAL(clicked()), this, SLOT(reject()));
+
+    // Generate label for proposal
+    ui->label->setTextFormat(Qt::RichText);
+    ui->label->setText(QString::fromStdString("You can now view your proposal in the "
+                                                             "core wallet or <a href=\"https://www.navexplorer.com/community-fund/proposal/"
+                                              + hash.ToString() + "\">on the navexplorer</a>"));
```
marcus290: actually I have one more suggestion - the success dialog box appears immediately after creating a proposal or preq, it has a link to NAVexplorer but the proposal/preq needs to be broadcast into the next block so this link will be broken until the explorer updates. So I would suggest either remove the link or link to the general proposal/preq page rather than the specific hash

hdnsimpson: Good point, I'll make this change tomorrow

src/qt/forms/communityfundcreateproposaldialog.ui
```diff
@@ -0,0 +1,246 @@
...
+       </item>
+       <item>
+        <widget class="QLabel" name="labelWarning">
+         <property name="font">
+          <font>
+           <pointsize>8</pointsize>
+          </font>
+         </property>
+         <property name="text">
+          <string>By submitting the proposal a 50 NAV deduction will occur from your wallet </string>
```
aguycalled: ideally those consensus-related values should be automatically taken from Params().GetConsensus()

aguycalled:
everything looks quite good from what i've tested in devnet. i only miss a filter for seeing proposals/prequests in pending state for which I haven't voted yet. this should be the view I see when I click the open community fund button from the new proposals alert dialog.

PR #437, opened 2019-04-09
Disables spam protection while IBD

aguycalled:
This PR disables the header spam protection while initial block download, as it could cause false positives.


matt-auckland:
Do we not think people will be spammed during IBD?

craigmacgregor:
@matt-auckland people doing initial block download are not contributing to the network weight with staking, so it's not a likely attack vector since spamming them and knocking them off the network does not reduce the network weight making a 51% attack easier.

PR #442, opened 2019-04-15
Use chainActive.Tip() instead of pindexBestHeader

aguycalled:
This PR introduces the use of chainActive.Tip() instead of pindexBestHeader as a reference to the best header in some critical parts of the code.


craigmacgregor:
I'm testing this today, will merge if my tests are successful.

PR #443, opened 2019-04-23
Community Fund GUI date buffer increase

aguycalled:
This PR includes a fix for a bug which caused a wallet to crash when loading the community fund graphical interface using the 'All' filter.


craigmacgregor:
nice, i had this happen to me and was going to log this bug but you already beat me to it with a fix :)

craigmacgregor:
mine still crashes when i use the "all" selector on the community fund tab.
```
EXCEPTION: St12out_of_range
basic_string::erase: __pos (hwich is 10) > this->size() (which is 0)
navcoin in Runaway exception
```

It actually happens to me on a few of the tabs; All, Expired & Rejected

Crashes on Ubuntu 18.04.

Doesn't crash on OSX for some reason.

PR #452, opened 2019-04-29
Added a short bash command for dependencies to build-unix.md

mxaddict:
Added a quick bash command for users to run the dev tools setup via bash using curl (most unix machines have both curl and bash)

### DIFF
--- a/doc/build-unix.md
+++ b/doc/build-unix.md
@@ -7,7 +7,11 @@ Some notes on how to build NavCoin Core in Unix.
 
 ## Building in Ubuntu 18.04
 
-You can easily build the dependencies by running the [NavCoin dev tools script here](https://github.com/NAVCoin/navcoin-dev-tools/blob/master/ubuntu-18.04-navcoin-core-dev-setup.sh).
+You can easily build the dependencies by running the [NavCoin dev tools script](https://github.com/NAVCoin/navcoin-dev-tools/blob/master/ubuntu-18.04-navcoin-core-dev-setup.sh) using the command bellow.
+
+```bash
+curl -o- https://raw.githubusercontent.com/NAVCoin/navcoin-dev-tools/master/ubuntu-18.04-navcoin-core-dev-setup.sh | bash
+```
 
 From the navcoin-core directory, you will still need to:
 ```bash


aguycalled:
Great, just realised a few lines down `./configure` is missing `-prefix=...`, we might use this PR to fix that too.

mxaddict:
> Great, just realised a few lines down `./configure` is missing `-prefix=...`, we might use this PR to fix that too.

Let me know what needs changing so I can add it.

aguycalled:
It’s necessary to indicate the compiler where to look for the built dependencies, otherwise it won’t use them.

https://github.com/NAVCoin/navcoin-core/blob/master/depends/README.md

Von meinem iPhone gesendet

mxaddict:
> It’s necessary to indicate the compiler where to look for the built dependencies, otherwise it won’t use them. https://github.com/NAVCoin/navcoin-core/blob/master/depends/README.md Von meinem iPhone gesendet
> […](#)
> Am 30.04.2019 um 15:19 schrieb Barry Deeney ***@***.***>: Great, just realised a few lines down ./configure is missing -prefix=..., we might use this PR to fix that too. Let me know what needs changing so I can add it. — You are receiving this because you commented. Reply to this email directly, view it on GitHub, or mute the thread.

This was not the case for me, I did not have to point the ./configure command to my depends, it found the dependencies and used them.

Maybe this is only the case when specifying a HOST that is not the native machines host?

I guess we should just add a note to the build-unix.md to the depends README.

The note should be about building for other platforms via unix machine.

mxaddict:
Actually, come to think of it, I see there is already a section on the build-unix.md file here: https://github.com/mxaddict/navcoin-core/blob/patch-1/doc/build-unix.md#arm-cross-compilation

It shows the use of `--prefix=`

Is this what you had in mind?

aguycalled:
in my case I always had to use `--prefix=`
if the dependencies are found without it, it's very likely they are already installed in your system

mxaddict:
Ahh, so the configure file is using my system versions of the dependencies.

That might explain why my GUI is themed the same way as my DE (colors and
fonts I mean)

PR #453, opened 2019-04-29
Currency Change now updates staking info as well

mxaddict:
When updating the currency setting on bottom right corner of the GUI wallet, the staking info section is now also updated instead of having to restart the wallet to see the change.


aguycalled:
utACK

PR #450, opened 2019-04-29
This PR is a combined fix for issues #445 and #446

mxaddict:
It seems that PR#447 did not fix the issue entirely, but it did point to the right direction.

This PR has changes that fix both #445 and #446

1. With original code on 4.6.0 RC, the issue was that a wallet with only a staking address would report negative values for staking report numbers, SAMPLE BELLOW:

```
{
  "2019-04-26 16:00:00": "0.00",
  "2019-04-25 16:00:00": "-8833.93873429",
  "2019-04-24 16:00:00": "-16719.44565578",
  "2019-04-23 16:00:00": "0.00",
  "2019-04-22 16:00:00": "2.00",
  "2019-04-21 16:00:00": "2.00",
  "2019-04-20 16:00:00": "2.00",
  "2019-04-19 16:00:00": "4.00",
  "2019-04-18 16:00:00": "2.00",
  "2019-04-17 16:00:00": "2.00",
  "2019-04-16 16:00:00": "4.00",
  "2019-04-15 16:00:00": "4.00",
  "2019-04-14 16:00:00": "4.00",
  "2019-04-13 16:00:00": "6.00",
  "2019-04-12 16:00:00": "4.00",
  "2019-04-11 16:00:00": "2.00",
  "2019-04-10 16:00:00": "2.00",
  "2019-04-09 16:00:00": "2.00",
  "2019-04-08 16:00:00": "6.00",
  "2019-04-07 16:00:00": "0.00",
  "2019-04-06 16:00:00": "0.00",
  "2019-04-05 16:00:00": "0.00",
  "2019-04-04 16:00:00": "6.00",
  "2019-04-03 16:00:00": "0.00",
  "2019-04-02 16:00:00": "0.00",
  "2019-04-01 16:00:00": "2.00",
  "2019-03-31 16:00:00": "2.00",
  "2019-03-30 16:00:00": "0.00",
  "2019-03-29 16:00:00": "0.00",
  "2019-03-28 16:00:00": "4.00",
  "Last 24H": "-8833.93873429",
  "Last 7 Days": "-25543.38439007",
  "Last 30 Days": "-25487.38426307",
  "Last 365 Days": "-25202.22851166",
  "Last All": "-25193.35322832",
  "Latest Stake": "-6046.0209064",
  "Latest Time": "2019-04-26 11:57:36",
  "Stake counted": 239,
  "time took (ms)": 1
}
```

[...]

PR #456, opened 2019-04-30
Fix for Payment Request reorganizations

aguycalled:
This PR prevents payment requests with invalid hashes (not set yet or out of the main chain) to count for the already requested balance of a proposal.

### DIFF
--- a/qa/rpc-tests/cfund-paymentrequest-extract-funds.py
+++ b/qa/rpc-tests/cfund-paymentrequest-extract-funds.py
@@ -47,15 +47,21 @@ class CommunityFundPaymentRequestExtractFundsTest(NavCoinTestFramework):
 
         # Create 6 payment requests
         paymentRequests = []
-        for x in range(6):
+        for x in range(5):
             paymentRequests.append(self.nodes[0].createpaymentrequest(proposalid0, 20, "test0")["hash"])
+
+	# The last 1 should be rejected as it excedes proposal balance
+        try:
+            paymentRequests.append(self.nodes[0].createpaymentrequest(proposalid0, 20, "test0")["hash"])
+            raise ValueError("Error should be thrown for invalid prequest")
+        except JSONRPCException as e:
+            assert("The transaction was rejected" in e.error['message'])
 
         slow_gen(self.nodes[0], 1)
 
         # One of them should have been rejected at creation
         valid = 0
         invalid = 0
-        invalid_pos = -1
 
         for paymentReq in paymentRequests:
             try:
@@ -64,13 +70,10 @@ class CommunityFundPaymentRequestExtractFundsTest(NavCoinTestFramework):
                 valid = valid + 1
             except JSONRPCException:
                 invalid = invalid + 1
-                invalid_pos = valid 
                 continue
 
         assert(valid == 5)
-        assert(invalid == 1)
-
-        paymentRequests.pop(invalid_pos)
+        assert(invalid == 0)
 
         assert(self.nodes[0].cfundstats()["funds"]["locked"] == locked_accepted)
         
--- a/qa/rpc-tests/cfund-paymentrequest-state-reorg.py
+++ b/qa/rpc-tests/cfund-paymentrequest-state-reorg.py
@@ -91,12 +91,9 @@ class CFundPaymentRequestStateReorg(NavCoinTestFramework):
 
         node_0_best_hash = self.nodes[0].getblockhash(self.nodes[0].getblockcount())
         node_1_best_hash = self.nodes[1].getblockhash(self.nodes[1].getblockcount())
-        node_0_chainwork = int(self.nodes[0].getblock(node_0_best_hash)["chainwork"], 16)
-        node_1_chainwork = int(self.nodes[1].getblock(node_1_best_hash)["chainwork"], 16)
 
-        # Node 0 and Node 1 have forked. Difficulty in Node 0 must be higher (hence lower nbits)
+        # Node 0 and Node 1 have forked. Difficulty in Node 0 must be higher (hence lower chainwork)
         assert(node_0_best_hash != node_1_best_hash)
-        assert(node_0_chainwork < node_1_chainwork)
 
         connect_nodes(self.nodes[0], 1)
 
--- a/qa/rpc-tests/cfund-paymentrequest-state-reorg.py
+++ b/qa/rpc-tests/cfund-paymentrequest-state-reorg.py
@@ -91,12 +91,12 @@ class CFundPaymentRequestStateReorg(NavCoinTestFramework):
 
         node_0_best_hash = self.nodes[0].getblockhash(self.nodes[0].getblockcount())
         node_1_best_hash = self.nodes[1].getblockhash(self.nodes[1].getblockcount())
-        node_0_nbits = int(self.nodes[0].getblock(node_0_best_hash)["bits"], 16)
-        node_1_nbits = int(self.nodes[1].getblock(node_1_best_hash)["bits"], 16)
+        node_0_chainwork = int(self.nodes[0].getblock(node_0_best_hash)["chainwork"], 16)
+        node_1_chainwork = int(self.nodes[1].getblock(node_1_best_hash)["chainwork"], 16)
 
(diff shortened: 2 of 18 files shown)


mxaddict:
@aguycalled  I see that your travis-ci run for this PR on x86_64 linux is failing with similar error to my test, any ideas on what the cause is?

qa/rpc-tests/cfund-paymentrequest-extract-funds.py
```diff
@@ -47,15 +47,21 @@ def run_test(self):
 
         # Create 6 payment requests
```
mxaddict: I think this should be `5` instead of `6`

aguycalled: it creates 6, 5 correct and the 6th should be rejected

src/main.cpp
```diff
@@ -3246,7 +3246,7 @@ bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pin
                 } else if (out.scriptPubKey.IsPayToPublicKey() || out.scriptPubKey.IsColdStaking()) {
                     uint160 hashBytes;
                     int type = 0;
-		    CTxDestination destination;
```
mxaddict: Maybe we should add `.editorconfig` to the project?

mxaddict: So we can standardize the white space and some other basic configs

aguycalled: +1

aguycalled:
The issue this PR fixes can be reproduced as follows:

> NODE1> staking false
NODE2> staking false
NODE2> addnode 127.0.0.1:node1port onetry
NODE1> generate 300
NODE1> donatefund 100
NODE1> getnewaddress
Save output in %1
NODE2> getnewaddress
Save output in %2
NODE1> sendtoaddress %2 5000000
NODE1> createproposal %1 100 3600 test
Save hash in %3
NODE1> generate 1
NODE2> listproposals
Assert output.size > 0
NODE1> proposalvote %3 yes
NODE1> generate 250
NODE1> createpaymentrequest %3 100 test true
Save output in %4
NODE2> disconnectnode 127.0.0.1:node1port 
NODE1> forcetransactions [“%4”]
NODE2> forcetransactions [“%4”]
NODE1> staking true
Wait for 3+ blocks.
Store best block hash in %5
NODE1> staking false
NODE2> staking true
Wait for 1 block. Node 2 only has one output so it will only stake once.
NODE2> staking false
NODE2> addnode 127.0.0.1:node1port onetry
Check that Node2.bestblockhash == %5

Tested succesfully with 2 nodes locally in devnet.
RPC test unit needs to be added.

craigmacgregor:
compiling and starting to review this now.

mxaddict:
> compiling and starting to review this now.

Looks like the latest build has an issue with `cfund-paymentrequest-state-reorg.py`

qa/rpc-tests/cfund-paymentrequest-state-reorg.py
```diff
@@ -0,0 +1,111 @@
...
+        self.nodes[1].staking(False)
+
+        node_0_best_hash = self.nodes[0].getblockhash(self.nodes[0].getblockcount())
+        node_1_best_hash = self.nodes[1].getblockhash(self.nodes[1].getblockcount())
+        node_0_nbits = int(self.nodes[0].getblock(node_0_best_hash)["bits"], 16)
+        node_1_nbits = int(self.nodes[1].getblock(node_1_best_hash)["bits"], 16)
+
+        # Node 0 and Node 1 have forked. Difficulty in Node 0 must be higher (hence lower nbits)
+        assert(node_0_best_hash != node_1_best_hash)
+        assert(node_0_nbits < node_1_nbits)
```
craigmacgregor: This assertion is failing because the nbits are the same. If i comment out this assertion the rest of the tests pass. So the nodes start with different best block hashes and then they are the same. 

Seems to have started failing after merging #450 and #459 but i don't really know why that would effect this test.

Is the nbit check required?

mxaddict: This means that some how the 2 nodes have the same best hash, even when they were staking while disconnected from each other.

How would that be possible?

mxaddict: > This means that some how the 2 nodes have the same best hash, even when they were staking while disconnected from each other.
> 
> How would that be possible?

Scratch that, so they best hash are not the same, but the node_0_nbits assert is failing?

mxaddict: Maybe this line should be :

```python
assert(node_0_nbits != node_1_nbits)
```

mxaddict: > Maybe this line should be :
> 
> ```python
> assert(node_0_nbits != node_1_nbits)
> ```

The nbits might actually be the same, because of this change:

https://github.com/mxaddict/navcoin-core/blob/bc692d86e2360ed5ef10bc5d6cb1de3c3385bb88/src/chainparams.cpp#L469

craigmacgregor: Yeah, the hashes are different but the nbits are the same. im reverting the target time to 30 seconds and seeing if it makes a difference.

mxaddict:
Roger that.

qa/rpc-tests/cfund-paymentrequest-state-reorg.py
craigmacgregor: Good thinking @mxaddict the test passes with the nTargetSpacing=30.

So i would assume its okay to remove the assertion that the nbits are different after reducing the devnet blocktime to 5s since it seems its just a function of having such a short block time.

Will wait for @aguycalled to confirm its okay to remove the assertion then I'm happy with this PR.

craigmacgregor:
> Roger that.
You mean ACK :P

mxaddict:
> > Roger that.
> 
> You mean ACK :P

I see `ACK` on alot of opensource projects, what does it actually stand for? :)

craigmacgregor:
> > > Roger that.
> > 
> > 
> > You mean ACK :P
> 
> I see `ACK` on alot of opensource projects, what does it actually stand for? :)

It's just short for "Acknowledged" AFAIK.

https://github.com/bitcoin/bitcoin/issues/6100

aguycalled:
I've changed the check so it looks at chainwork instead of nbits.
chainwork = prevblock.chainwork + nbits

mxaddict:
Looks like final commit fixed the issue with the test.

@craigmacgregor ready to merge? :)

qa/rpc-tests/cfund-paymentrequest-state-reorg.py
```diff
@@ -91,12 +91,9 @@ def run_test(self):
 
         node_0_best_hash = self.nodes[0].getblockhash(self.nodes[0].getblockcount())
         node_1_best_hash = self.nodes[1].getblockhash(self.nodes[1].getblockcount())
-        node_0_nbits = int(self.nodes[0].getblock(node_0_best_hash)["bits"], 16)
-        node_1_nbits = int(self.nodes[1].getblock(node_1_best_hash)["bits"], 16)
 
-        # Node 0 and Node 1 have forked. Difficulty in Node 0 must be higher (hence lower nbits)
+        # Node 0 and Node 1 have forked. Difficulty in Node 0 must be higher (hence lower chainwork)
```
mxaddict: I think this should be: 
```
       # Node 0 and Node 1 have forked. Difficulty in Node 0 must be higher
```

qa/rpc-tests/cfund-paymentrequest-state-reorg.py
aguycalled: opted for remove that check

craigmacgregor:
Travis failing again. Investigating.

aguycalled:
it's failing again uniquely because of a random fee in the coldstaking_* tests

craigmacgregor:
my local build and tests pass in Ubuntu 18.04

PR #459, opened 2019-05-01
Missing increased buffer for cfund gui

aguycalled:
Pull request #443 was missing an increased buffer in communityfunddisplay.cpp which caused the wallet to crash when using the CFund GUI on some systems. This pull requests corrects the size.


craigmacgregor:
Tested ACK

PR #463, opened 2019-05-02
Updated all references to old Repo Address for NavCoin github current url

mxaddict:
Updated the gitlian-linux.yml Repo Address for NavCoin

I think this only works because github does auto redirect of old repo URL?


aguycalled:
yes, navcoindev is the old account but i guess craig linked them. approved for merge

mxaddict:
I think that it should match the repo url exactly as RFC for urls is case
sensitive.

I guess if we want we could rename the repo to lowercase as it is better
looking as a URL.

aguycalled:
yes, i would rather rename the url @craigmacgregor

craigmacgregor:
repo URL has been renamed to `navcoin` all lower case. please update PR

mxaddict:
> organisation URL has been renamed to `navcoin` all lower case. please update PR

I've updated the PR

craigmacgregor:
approved, re-running travis then will merge when it passes.

craigmacgregor:
`The job exceeded the maximum time limit for jobs, and has been terminated.`

:/

PR #464, opened 2019-05-02
Adds OP_POOL to scripts

dantudor:
When blocks are created by the NavPool they will create an additional 0 value transaction on the coinbase. This is to signal that the block has been created by a pooled resource.

The new op code is `OP_POOL` with a value of `0xd0`

It will use the transaction label `TX_POOL` with a value of `pool_staking`


aguycalled:
ACK approval pending travis

PR #466, opened 2019-05-04
Added error log tab in debug window

mxaddict:
Added error log tab in debug window that parses last 500 debug.log entries for errors and loads more errors as they come in while debugwindow is open


mxaddict:
@aguycalled @craigmacgregor  This PR is ready for review, I have added the new `error.log` support and optimized the `Debug Window` code to not lock up the wallet.

aguycalled:
I'm not sure if filtering which entries should go to the error log by trying to find a string ('ERROR') is the right approach.

Using error() vs LogPrint* to decide on where should it be logged looks like a more efficient, elegant and future-prone solution.

Thoughts?

mxaddict:
> I'm not sure if filtering which entries should go to the error log by trying to find a string ('ERROR') is the right approach.
> 
> Using error() vs LogPrint* to decide on where should it be logged looks like a more efficient, elegant and future-prone solution.
> 
> Thoughts?

Agreed, I will make the change now.

mxaddict:
> I'm not sure if filtering which entries should go to the error log by trying to find a string ('ERROR') is the right approach.
> 
> Using error() vs LogPrint* to decide on where should it be logged looks like a more efficient, elegant and future-prone solution.
> 
> Thoughts?

I've pushed the changes, waiting on travis :smile:

craigmacgregor:
Is there some way to generate errors which will display in the error tab so i can test this is working?

mxaddict:
> Is there some way to generate errors which will display in the error tab so i can test this is working?

I just add a error("test") call to the init, compiled then checked the error log :D

mxaddict:
@aguycalled were you able to test a node running this?

proletesseract:
approved, waiting for travis to complete then will merge

mxaddict:
Looks like changes to QT version on master branch break this PR, I will fix the branch now

mxaddict:
@proletesseract  I've made the changes to fix the QT dependency issue, Build should pass now :+1:

mxaddict:
@proletesseract tests passed, merging :+1:

PR #469, opened 2019-05-06
Added some new FIAT currencies and enabled auto price update (Default value 300K ms)

mxaddict:
Added some new FIAT currencies to the QT wallet.

### DIFF
--- a/src/qt/forms/optionsdialog.ui
+++ b/src/qt/forms/optionsdialog.ui
@@ -670,6 +670,12 @@
          </item>
          <item>
           <widget class="QValueComboBox" name="unit">
+           <property name="editable">
+            <bool>true</bool>
+           </property>
+           <property name="insertPolicy">
+            <enum>QComboBox::NoInsert</enum>
+           </property>
            <property name="toolTip">
             <string>Choose the default subdivision unit to show in the interface and when sending coins.</string>
            </property>
--- a/src/qt/guiconstants.h
+++ b/src/qt/guiconstants.h
@@ -5,6 +5,9 @@
 #ifndef NAVCOIN_QT_GUICONSTANTS_H
 #define NAVCOIN_QT_GUICONSTANTS_H
 
+/* Milliseconds between price updates */
+static const int PRICE_UPDATE_DELAY = 300000;
+
 /* Milliseconds between model updates */
 static const int MODEL_UPDATE_DELAY = 250;
 
--- a/src/qt/navcoingui.cpp
+++ b/src/qt/navcoingui.cpp
@@ -250,7 +250,14 @@ NavCoinGUI::NavCoinGUI(const PlatformStyle *platformStyle, const NetworkStyle *n
     QHBoxLayout *frameBlocksLayout = new QHBoxLayout(frameBlocks);
     frameBlocksLayout->setContentsMargins(3,0,3,0);
     frameBlocksLayout->setSpacing(3);
-    unitDisplayControl = new UnitDisplayStatusBarControl(platformStyle);
+    unitDisplayControl = new QComboBox();
+    unitDisplayControl->setEditable(true);
+    unitDisplayControl->setInsertPolicy(QComboBox::NoInsert);
+    Q_FOREACH(NavCoinUnits::Unit u, NavCoinUnits::availableUnits())
+    {
+        unitDisplayControl->addItem(QString(NavCoinUnits::name(u)), u);
+    }
+    connect(unitDisplayControl,SIGNAL(currentIndexChanged(int)),this,SLOT(comboBoxChanged(int)));
     labelEncryptionIcon = new QLabel();
     labelStakingIcon = new QLabel();
     labelPrice = new QLabel();
@@ -585,14 +592,6 @@ void NavCoinGUI::createMenuBar()
         settings->addSeparator();
         settings->addAction(toggleStakingAction);
         settings->addSeparator();
-        QMenu* currency = settings->addMenu( tr("Currency") );
-        Q_FOREACH(NavCoinUnits::Unit u, NavCoinUnits::availableUnits())
-        {
-            QAction *menuAction = new QAction(QString(NavCoinUnits::name(u)), this);
-            menuAction->setData(QVariant(u));
-            currency->addAction(menuAction);
-        }
-        connect(currency,SIGNAL(triggered(QAction*)),this,SLOT(onCurrencySelection(QAction*)));
         settings->addAction(updatePriceAction);
     }
     settings->addAction(optionsAction);
@@ -607,14 +606,6 @@ void NavCoinGUI::createMenuBar()
     help->addAction(aboutAction);
     help->addAction(infoAction);
     help->addAction(aboutQtAction);
-}
-
-void NavCoinGUI::onCurrencySelection(QAction* action)
-{
-    if (action)
-    {
-        clientModel->getOptionsModel()->setDisplayUnit(action->data());
-    }
 }
 
 void NavCoinGUI::createToolBars()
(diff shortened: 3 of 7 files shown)


aguycalled:
this should be thoroughly tested. when it was enabled before it caused the wallet to eventually lose network connections after long periods of use because the sockets descriptors were not correctly released after each request

mxaddict:
> this should be thoroughly tested. when it was enabled before it caused the wallet to eventually lose network connections after long periods of use because the sockets descriptors were not correctly released after each request

Is that a QT related issue? I'll see if I can change it to release the connections

Cause the network calls are using the QT libs

mxaddict:
I'll compile a version that updates prices every second, and see if it looses connection after a few days.

aguycalled:
id say deletelater() should release them, but we were seeing that side effect before

mxaddict:
> id say deletelater() should release them, but we were seeing that side effect before

Possibly a bug with the QT 5.7.1 libs? I'll build against both 5.7.1 and 5.9.8 and see if the issue is fixed.

mxaddict:
> id say deletelater() should release them, but we were seeing that side effect before

Looks like I got it to happen, is this the issue? :

```
2019-05-06 19:16:18 socket send error Bad file descriptor (9)
2019-05-06 19:16:19 INFO: Updating prices
2019-05-06 19:16:19 ERROR: Could not parse price data json 'boost::property_tree::json_parser::json_parser_error'
2019-05-06 19:16:24 INFO: Updating prices
```

node crashes/closes after the `socket send error Bad file descriptor` error

Which I think comes from here:
https://github.com/NAVCoin/navcoin-core/blob/59a4ba7486e886c73196b6741b94b229c0d02503/src/net.cpp#L846-L854

mxaddict:
> > id say deletelater() should release them, but we were seeing that side effect before
> 
> Looks like I got it to happen, is this the issue? :
> 
> ```
> 2019-05-06 19:16:18 socket send error Bad file descriptor (9)
> 2019-05-06 19:16:19 INFO: Updating prices
> 2019-05-06 19:16:19 ERROR: Could not parse price data json 'boost::property_tree::json_parser::json_parser_error'
> 2019-05-06 19:16:24 INFO: Updating prices
> ```
> 
> node crashes/closes after the `socket send error Bad file descriptor` error
> 
> Which I think comes from here:
> https://github.com/NAVCoin/navcoin-core/blob/59a4ba7486e886c73196b6741b94b229c0d02503/src/net.cpp#L846-L854

Seems this was not actually related to the current issue

mxaddict:
I was able to replicate the issue by setting up a localhost http server to serve the JSON file

And updated the price check interval to 100ms

Node built with QT 5.7.1 fails in about 10 secs

Node built with QT 5.9.8 failes in about 16 minutes

both with same error:
```
QThreadPipe: Unable to create pipe: Too many open files
```

mxaddict:
So I think this behavior should be normal.

My guess is since I was testing with 100ms interval, the QT code that would handle deleteLater() could not keep up with the new requests

So in a more realistic setting were lets say the interval is 5 minutes, this should not be an issue.

@aguycalled @craigmacgregor  do you agree?

mxaddict:
I think this PR is ready for review :smile:

aguycalled:
Has it been confirmed through testing the issue won't happen in a real world scenario with a higher interval, or is it assumed? We saw the issue before with a similar interval.

mxaddict:
> Has it been confirmed through testing the issue won't happen in a real world scenario with a higher interval, or is it assumed? We saw the issue before with a similar interval.

It's assumed as if now (I can't time travel, sorry)

I'll still running the node with the 5 minute interval and will leave it running for a few days to confirm that it does not loose a connection / crash

But I'm almost :100: % certain that the issue was fixed by moving this line:

```cpp
reply->deleteLater();
```

mxaddict:
I've done additional testing and found that the wallet indeed does not release the socket files in `/proc/<PID>/fd` even after the request is run

I'll try to replace the HTTP request code with libcurl instead of using QT5's classes and see if it solves the issue

Sample filesi n `/proc/<PID>/fd` for a node running at 3 sec update interval:
```
lrwx------ 1 mxaddict mxaddict 64 May  7 15:42 92 -> 'socket:[759652]'
lrwx------ 1 mxaddict mxaddict 64 May  7 15:42 93 -> 'socket:[765795]'
lrwx------ 1 mxaddict mxaddict 64 May  7 15:42 94 -> 'socket:[765803]'
lrwx------ 1 mxaddict mxaddict 64 May  7 15:42 95 -> 'socket:[763781]'
lrwx------ 1 mxaddict mxaddict 64 May  7 15:42 97 -> 'socket:[767521]'
lrwx------ 1 mxaddict mxaddict 64 May  7 15:42 96 -> 'anon_inode:[eventfd]'
lrwx------ 1 mxaddict mxaddict 64 May  7 15:42 98 -> 'socket:[767572]'
lrwx------ 1 mxaddict mxaddict 64 May  7 15:42 99 -> 'socket:[768311]'
lrwx------ 1 mxaddict mxaddict 64 May  7 15:42 101 -> 'socket:[767611]'
lrwx------ 1 mxaddict mxaddict 64 May  7 15:42 100 -> 'socket:[772218]'
lrwx------ 1 mxaddict mxaddict 64 May  7 15:42 103 -> 'socket:[766455]'
lrwx------ 1 mxaddict mxaddict 64 May  7 15:42 102 -> 'socket:[767633]'
lrwx------ 1 mxaddict mxaddict 64 May  7 15:42 104 -> 'socket:[764853]'
lrwx------ 1 mxaddict mxaddict 64 May  7 15:42 105 -> 'socket:[770109]'
lrwx------ 1 mxaddict mxaddict 64 May  7 15:42 106 -> 'socket:[766510]'
lrwx------ 1 mxaddict mxaddict 64 May  7 15:42 109 -> 'socket:[773138]'
lrwx------ 1 mxaddict mxaddict 64 May  7 15:42 108 -> 'socket:[773124]'
lrwx------ 1 mxaddict mxaddict 64 May  7 15:42 107 -> 'socket:[770134]'
lrwx------ 1 mxaddict mxaddict 64 May  7 15:43 110 -> 'socket:[768550]'
lrwx------ 1 mxaddict mxaddict 64 May  7 15:43 111 -> 'socket:[768576]'
```

The HTTP request seem to be the files with `socket:[<int>]` names

mxaddict:
@aguycalled @craigmacgregor  latest commit in this PR replaces QT's network manager with a plain old `curl` + `std::thread` setup.

I can confirm that this implementation was properly releasing the sockets after request was done.

aguycalled:
utACK

mxaddict:
@aguycalled I've had a node running these changes for a week, seems to be working and not holding the sockets.

mxaddict:
output

@aguycalled The currencies list is getting kinda bloated, maybe I should replace those menu's with a searchable dropdown? Show top 5, then have a search bar?

aguycalled:
> output
> 
> @aguycalled The currencies list is getting kinda bloated, maybe I should replace those menu's with a searchable dropdown? Show top 5, then have a search bar?

ACK

aguycalled:
Had a node running for 24h+ on Debian

mxaddict:
I'm adding the GUI change for the coin list ( then you guys can review )

mxaddict:
* I removed the currency picker in Settings->Currency (Was pretty ugly and redundant)
* Updated currency menu in Options Dialog to be a combobox
* Updated currency menu in bottom right of wallet ui to be a combobox

mxaddict:
Rebased the branch from master

mxaddict:
@aguycalled I wanna change my bounty address to `XakMNvZREj5XWUTnLJWMqdU62vcoP3ksDY3HAoAEjtKhGLyUSakjLrMjgCuJN` for future bounties

mxaddict:
Rebased from master

mxaddict:
Demo:

output

proletesseract:
building now.

mxaddict:
@aguycalled you missed a bounty :rofl:

aguycalled:
Paid a07b6e5541ee6c9a249fb59020d573d56bc0a8de26ee853dfde823df68673b79

Issue #457, opened 2019-05-01
CFund GUI requires wallet unlock to create proposal / payment request.


craigmacgregor:
Using the new CFund GUI in 4.6.0-rc to create a payment request or proposal requires the user to unlock the wallet. There is no dialog to enter the passphrase to unlock the wallet and create the transaction. There is only an error alert dialog. 

<img width="441" alt="Screen Shot 2019-05-01 at 3 50 56 PM" src="https://user-images.githubusercontent.com/6902746/57005093-f7006280-6c28-11e9-83f4-2c0a9551247f.png">

I had to go to the debug window and use the walletpassphrase command to unlock my wallet before I was able to create the proposal.

Ideally the flow would be the same as creating a regular transaction. The user would be prompted to enter their password and the wallet would only remain unlocked for the duration of creating and signing the transaction then immediately revert to being locked again.

Steps to reproduce;

- Open navcoin-qt
- Encrypt wallet
- Navigate to Community Fund tab
- Try to create a proposal or a payment request

craigmacgregor:
@aguycalled do you want to allocate a bounty to this?

francisjyap:
@aguycalled 

NVRFaYW5PFJFG7xQomPqczjZFSVXAbqh2E

aguycalled:
paid 144b6197e6c3b8e0fb3c1e28d430b44db09ff9f8b484c8186474c6d78f7efbcb

PR #474, opened 2019-05-09
Added password prompt for CFund payment/proposal dialogs

francisjyap:
Added password prompt for CFund payment/proposal dialogs

This is for issue #457

### DIFF
--- a/src/qt/communityfundcreatepaymentrequestdialog.cpp
+++ b/src/qt/communityfundcreatepaymentrequestdialog.cpp
@@ -1,19 +1,21 @@
 #include "communityfundcreatepaymentrequestdialog.h"
-#include "ui_communityfundcreatepaymentrequestdialog.h"
 #include "communityfundsuccessdialog.h"
 #include "sendcommunityfunddialog.h"
+#include "ui_communityfundcreatepaymentrequestdialog.h"
 
 #include <QMessageBox>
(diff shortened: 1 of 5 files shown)


mxaddict:
utACK

mxaddict:
Built on Ubuntu 18.04 (I'll do a testnet/devnet test for the functionality)

aguycalled:
ACK tested on debian

aguycalled:
@francisjyap please leave a comment in the original issue with a navcoin address for payment

mxaddict:
@aguycalled the test seems to keep failing (Timeout) Any ideas how to get the travis build to work?

aguycalled:
use travis_wait to extend the test job max execution time

aguycalled:
Looks like it is not possible to extend the job duration to more than 50 min for public repositories using the free version of travis. Any alternatives different from reducing our test units?

mxaddict:
> Looks like it is not possible to extend the job duration to more than 50 min for public repositories using the free version of travis. Any alternatives different from reducing our test units?

I think if we adjust testnet/devnet to so that the tests pass faster?

Like for example, we can have the static staking reward take only 150 (50 per step)  blocks instead of the current 300 (100 per step)

And I think the same can be done for cfund tests as well

Issue #193, opened 2018-04-13
Importaddress in the console doesn´t seem to return correct amounts


sergiohache:
The command importaddress in the console doesn´t seem to return updated or correct amounts, there could be a lag or the calculation was not made correctly. This was reproduced yesterday in the testnet with v-4.1.2 cold staking between staking and spending wallets.

[](Elegantt!UNITO-UNDERSCORE!data:dont!UNITO-UNDERSCORE!delete{"ignored":false,"autoPlanned":true,"ownerId":false,"dependencies":[],"psd":"2018-04-16T09:00:00+12:00","ped":"2018-04-20T18:00:00+12:00"})

PR #479, opened 2019-05-10
Fix for importaddress returning incorrect amounts

francisjyap:
Fix for importaddress returning incorrect amounts

This is for Issue #193


aguycalled:
Maybe the issue is not clear enough, but I think he refers to an issue when importing cold staking addresses.

francisjyap:
@aguycalled Updated the test case to include cold staking addresses.

aguycalled:
just tested it, importing the spending or staking address in the second node does not show the balance as if it was an address owned by the wallet.

francisjyap:
@aguycalled oh, i assumed that the balance from addresses imported were not to be shown as if it was owned by the wallet, but only show transactions on the watched address. In this case, should commands like getbalance be updated to include balances from watched addresses?

aguycalled:
you are right, only the transactions are watched, but those do not appear for spending or staking addresses.

mxaddict:
> you are right, only the transactions are watched, but those do not appear for spending or staking addresses.

So does that mean:

```
ZZZ = normal address # Has balance
XXX = staking # No balance
YYY = spending # No balance
[YYY,XXX] = cold address # Has balance

Import ZZZ = works (You see transactions + correct amount)
Import XXX = (does this work?)
Import YYY = (does this work?)
Import [YYY,XXX] = works (You see transactions + correct amount)
```

aguycalled:
> > you are right, only the transactions are watched, but those do not appear for spending or staking addresses.
> 
> So does that mean:
> 
> ```
> ZZZ = normal address # Has balance
> XXX = staking # No balance
> YYY = spending # No balance
> [YYY,XXX] = cold address # Has balance
> 
> Import ZZZ = works (You see transactions + correct amount)
> Import XXX = (does this work?)
> Import YYY = (does this work?)
> Import [YYY,XXX] = works (You see transactions + correct amount)
> ```

yes

mxaddict:
@francisjyap I think you just need to add those scenarios to the functional test, then fix the issue :+1: if it still exists.

@aguycalled  so the balance is not supposed to update, only the transactions list correct?

aguycalled:
watchonly address balance is only shown in qt iirc

francisjyap:
@mxaddict @aguycalled I updated the test cases per your suggestions and the transactions and balance from the imported addresses are reflected correctly.

I added a Watched label on the UI showing the total balance from watched addresses. The Total label is also updated to reflect the total balance plus the total watched balance.

mxaddict:
utACK

qa/rpc-tests/importaddress.py
```diff
@@ -0,0 +1,71 @@
...
+        assert_equal(512, transactions[0]['amount'])
+        assert_equal(128, transactions[1]['amount'])
+        assert_equal(256, transactions[2]['amount'])
+
+        # Assert total balance
+        assert_equal(896, self.nodes[2].getbalance("*", 1, True))
+
+
+if __name__ == '__main__':
+    GetStakeReport().main()
```
mxaddict: Needs to match the name of the file

qa/rpc-tests/importaddress.py
```diff
@@ -0,0 +1,71 @@
...
+        self.nodes[0].sendtoaddress(coldstaking_address, 256)
+        self.nodes[0].sendtoaddress(coldstaking_address, 128)
+        self.nodes[0].generate(6)
+        self.sync_all()
+
+        # Import address with balances to node[2]
+        self.nodes[2].importaddress(address)
+        self.nodes[2].importaddress(spending_address)
+        self.nodes[2].importaddress(staking_address)
+        self.nodes[2].importaddress(coldstaking_address)
```
aguycalled: if you import both the spending/staking and the cold staking address, you can't discern if the txs are seen because the wallet is watching the cold staking address (which was already working) or because it is watching the spending/staking addresses (which were not working before). I'd suggest a setup with one node for each case.

mxaddict:
@francisjyap any updates on this issue?

francisjyap:
@aguycalled @mxaddict So i found out what was causing the cold staking transactions from appearing when importing a staking/spending address. Cold staking txs would only compare against addresses that are owned by the wallet and not against watched addresses. I also updated the test case.

mxaddict:
utACK

aguycalled:
utACK

mxaddict:
Full test suite + new test cases for this scenario also pass

aguycalled:
Tested OSX

aguycalled:
paid 5ab73dece539ab5a29b196a4af43e1834c9a5e35da1c977b92d71a005fb4feed

PR #480, opened 2019-05-12
Fix RPC Tests

aguycalled:
This PR reduces the difficulty for devnet's genesis block and the cfund voting cycle length in order to reduce the total execution time of the RPC tests.


mxaddict:
utACK

proletesseract:
Compiled, running the test suite.

proletesseract:
all passed.

PR #485, opened 2019-05-16
Added new attributes to the `getstakinginfo` and `getstakereport` commands

mxaddict:
Added new attributes to the `getstakinginfo` and `getstakereport` commands

`getstakinginfo` now has `expecteddailyreward`

`getstakereport` now has `Last 7 Days Avg`, `Last 30 Days Avg` and `Last 365 Days Avg`

reports looks like these:
```json
{
  "enabled": true,
  "staking": true,
  "errors": "",
  "currentblocksize": 1000,
  "currentblocktx": 0,
  "difficulty": 537692.0995617752,
  "search-interval": 16,
  "weight": 1198583732549,
  "netstakeweight": 1339930863008790,
  "expectedtime": 33537,
  "expecteddailyreward": 5.15236448
}
```

```json
{
  "2019-05-15 16:00:00": "2.00",
  "2019-05-14 16:00:00": "2.00",
  "2019-05-13 16:00:00": "2.00",
  "2019-05-12 16:00:00": "2.00",
  "2019-05-11 16:00:00": "2.00",
  "2019-05-10 16:00:00": "6.0001",
  "2019-05-09 16:00:00": "4.00",
  "2019-05-08 16:00:00": "4.00",
  "2019-05-07 16:00:00": "6.00",
  "2019-05-06 16:00:00": "4.00",
  "2019-05-05 16:00:00": "2.00",
  "2019-05-04 16:00:00": "4.00",
  "2019-05-03 16:00:00": "8.00",
  "2019-05-02 16:00:00": "4.00",
  "2019-05-01 16:00:00": "6.00",
  "2019-04-30 16:00:00": "2.00",
  "2019-04-29 16:00:00": "0.00",
  "2019-04-28 16:00:00": "4.00",
  "2019-04-27 16:00:00": "4.00",
  "2019-04-26 16:00:00": "2.00",
  "2019-04-25 16:00:00": "4.00",
  "2019-04-24 16:00:00": "4.00",
  "2019-04-23 16:00:00": "0.00",
  "2019-04-22 16:00:00": "2.00",
  "2019-04-21 16:00:00": "2.00",
  "2019-04-20 16:00:00": "2.00",
  "2019-04-19 16:00:00": "4.00",
  "2019-04-18 16:00:00": "2.00",
  "2019-04-17 16:00:00": "2.00",
  "2019-04-16 16:00:00": "4.00",
  "Last 24H": "2.00",
  "Last 7 Days": "22.0001",
  "Last 7 Days Avg": "3.14287142",
  "Last 30 Days": "98.0001",
  "Last 30 Days Avg": "3.26667",
  "Last 365 Days": "425.2213209",
  "Last 365 Days Avg": "1.16498992",
  "Last All": "438.03126175",
  "Latest Stake": "2.00",
  "Latest Time": "2019-05-15 20:53:04",
  "Stake counted": 274,
  "time took (ms)": 1
}
```

### DIFF
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -3258,7 +3258,7 @@ vStakePeriodRange_T PrepareRangeForStakeReport()
     }
 
     // prepare subtotal range of last 24H, 1 week, 30 days, 1 years
-    int GroupDays[5][2] = { {1 ,0}, {7 ,0 }, {30, 0}, {365, 0}, {99999999, 0}};
+    int GroupDays[5][2] = { {1, 0}, {7, 0}, {30, 0}, {365, 0}, {99999999, 0}};
     std::string sGroupName[] = {"24H", "7 Days", "30 Days", "365 Days", "All" };
 
     nToday = GetTime();
@@ -3301,10 +3301,29 @@ UniValue getstakereport(const UniValue& params, bool fHelp)
 
     vStakePeriodRange_T::iterator vIt;
 
+    // Span of days to compute average over
+    int nDays = 0;
+
     // report it
     for(vIt = aRange.begin(); vIt != aRange.end(); vIt++)
     {
+        // Add it to results
         result.push_back(Pair(vIt->Name, FormatMoney(vIt->Total).c_str()));
+
+        // Get the nDays value
+        nDays = 0;
+        if (vIt->Name == "Last 7 Days")
+            nDays = 7;
+        else if (vIt->Name == "Last 30 Days")
+            nDays = 30;
+        else if (vIt->Name == "Last 365 Days")
+            nDays = 365;
+
+        // Check if we need to add the average
+        if (nDays > 0) {
+            // Add the Average
+            result.push_back(Pair(vIt->Name + " Avg", FormatMoney(vIt->Total / nDays).c_str()));
+        }
     }
 
     vIt--;
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -3186,9 +3186,9 @@ int64_t GetFirstStakeTime()
     const CWalletTx* tx;
 
     // scan the entire wallet transactions
-    for (auto it = pwalletMain->mapWallet.end(); it != pwalletMain->mapWallet.begin(); --it)
+    for (auto it = pwalletMain->wtxOrdered.begin(); it != pwalletMain->wtxOrdered.end(); ++it)
     {
-        tx = &(*it).second;
+        tx = (*it).second.first;
 
         // Check if we have a useable tx
         if (IsTxCountedAsStaked(tx))
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -34,6 +34,7 @@
 using namespace std;
 
 int64_t nWalletUnlockTime;
+int64_t nWalletFirstStakeTime = -1;
 static CCriticalSection cs_nWalletUnlockTime;
 Navtech navtech;
 
@@ -3182,6 +3183,10 @@ CAmount GetTxStakeAmount(const CWalletTx* tx)
 // Returns -1 (Zero) if has not staked yet
 int64_t GetFirstStakeTime()
 {
+    // Check if we already know when
+    if (nWalletFirstStakeTime > 0)
+        return nWalletFirstStakeTime;
+
     // Need a pointer for the tx
     const CWalletTx* tx;
 
@@ -3191,12 +3196,14 @@ int64_t GetFirstStakeTime()
         tx = (*it).second.first;
 
         // Check if we have a useable tx
-        if (IsTxCountedAsStaked(tx))
-            return tx->nTime;
+        if (IsTxCountedAsStaked(tx)) {
+            nWalletFirstStakeTime = tx->nTime; // Save it for later use
+            return nWalletFirstStakeTime;
+        }
     }
 
     // Did not find the first stake
-    return -1;
+    return nWalletFirstStakeTime;
 }
 
 // **em52: Get total coins staked on given period
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -3150,6 +3150,55 @@ struct StakePeriodRange_T {
 
 typedef std::vector<StakePeriodRange_T> vStakePeriodRange_T;
 
+// Check if we have a Tx that can be counted in staking report
+bool IsTxCountedAsStaked(const CWalletTx* tx)
+{
+    // orphan block or immature
+    if ((!tx->GetDepthInMainChain()) || (tx->GetBlocksToMaturity() > 0))
+        return false;
+
+    // abandoned transactions
+    if (tx->isAbandoned())
+        return false;
+
+    // transaction other than POS block
+    return tx->IsCoinStake();
+}
+
+// Get the amount for a staked tx used in staking report
+CAmount GetTxStakeAmount(const CWalletTx* tx)
+{
+    // use the cached amount if available
+    if ((tx->fCreditCached || tx->fColdStakingCreditCached) && (tx->fDebitCached || tx->fColdStakingDebitCached))
+        return tx->nCreditCached + tx->nColdStakingCreditCached - tx->nDebitCached - tx->nColdStakingDebitCached;
+    // Check for cold staking 
+    else if (tx->vout[1].scriptPubKey.IsColdStaking())
+        return tx->GetCredit(pwalletMain->IsMine(tx->vout[1])) - tx->GetDebit(pwalletMain->IsMine(tx->vout[1]));
+
+    return tx->GetCredit(ISMINE_SPENDABLE) + tx->GetCredit(ISMINE_STAKABLE) - tx->GetDebit(ISMINE_SPENDABLE) - tx->GetDebit(ISMINE_STAKABLE);
+}
+
+// Gets timestamp for first stake
+// Returns -1 (Zero) if has not staked yet
+int64_t GetFirstStakeTime()
+{
+    // Need a pointer for the tx
+    const CWalletTx* tx;
+
+    // scan the entire wallet transactions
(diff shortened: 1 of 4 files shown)


src/wallet/rpcwallet.cpp
```diff
@@ -3300,10 +3300,29 @@ UniValue getstakereport(const UniValue& params, bool fHelp)
...
+            nDays = 7;
+        else if (vIt->Name == "Last 30 Days")
+            nDays = 30;
+        else if (vIt->Name == "Last 365 Days")
+            nDays = 365;
+
+        // Check if we need to add the average
+        if (nDays > 0) {
+            // Add the Average
+            result.push_back(Pair(vIt->Name + " Avg", FormatMoney(vIt->Total / nDays).c_str()));
```
aguycalled: Shouldn't the average be calculated only over the amount for days where a stake happened? If my wallet is 60 days old, it'll need 305 days to show an accurate number in "365 Avg", for example.

mxaddict: So you mean calculate the average stake for the last X days where X Days is the number of days past since your first stake that is no older than 365 days?

Well, that would not really be a 365 day average, that would be more an average since staking.

The current formula would be 100% accurate if your wallet is 1 year old, but as you say it would be completely off if your wallet has only been staking for like 45 days.

I'll see what I can come up with for making the average more accurate.

I think I can calculate the wallet age based on first stake date and use it to adjust nDays value.

aguycalled: I would define “X Avg” as the daily average received as rewards over the last X days of active staking.

mxaddict:
Ahh, so just add a new attribute to the response?

aguycalled:
> Ahh, so just add a new attribute to the response?
> […](#)

what I mean is, if my wallet is only n days old “X Avg” should show Sum(stakes[today-X:today])/min(X,n)

mxaddict:
> > Ahh, so just add a new attribute to the response?
> > […](#)
> 
> what I mean is, as a general definition, “X Avg” should show Sum(stakes[today-X:today])/min(X,n) for a wallet which is n days old

You mean like this:
```json
{
  "Last 24H": "2.00",
  "Last 7 Days": "22.0001",
  "Last 7 Days Avg": "3.14287142",
  "Last 30 Days": "98.0001",
  "Last 30 Days Avg": "3.26667",
  "Last 365 Days": "425.2213209",
  "Last 365 Days Avg": "1.16498992",
  "Last X Avg": "3.14287142",
  "Last All": "438.03126175"
}
```

aguycalled:
> > > Ahh, so just add a new attribute to the response?
> > > […](#)
> > 
> > 
> > what I mean is, as a general definition, “X Avg” should show Sum(stakes[today-X:today])/min(X,n) for a wallet which is n days old
> 
> You mean like this:
> 
> ```json
> {
>   "Last 24H": "2.00",
>   "Last 7 Days": "22.0001",
>   "Last 7 Days Avg": "3.14287142",
>   "Last 30 Days": "98.0001",
>   "Last 30 Days Avg": "3.26667",
>   "Last 365 Days": "425.2213209",
>   "Last 365 Days Avg": "1.16498992",
>   "Last X Avg": "3.14287142",
>   "Last All": "438.03126175"
> }
> ```

Let X be 7, 30 or 365.

src/wallet/rpcwallet.cpp
```diff
@@ -3149,6 +3149,55 @@ struct StakePeriodRange_T {
...
+CAmount GetTxStakeAmount(const CWalletTx* tx)
+{
+    // use the cached amount if available
+    if ((tx->fCreditCached || tx->fColdStakingCreditCached) && (tx->fDebitCached || tx->fColdStakingDebitCached))
+        return tx->nCreditCached + tx->nColdStakingCreditCached - tx->nDebitCached - tx->nColdStakingDebitCached;
+    // Check for cold staking 
+    else if (tx->vout[1].scriptPubKey.IsColdStaking())
+        return tx->GetCredit(pwalletMain->IsMine(tx->vout[1])) - tx->GetDebit(pwalletMain->IsMine(tx->vout[1]));
+
+    return tx->GetCredit(ISMINE_SPENDABLE) + tx->GetCredit(ISMINE_STAKABLE) - tx->GetDebit(ISMINE_SPENDABLE) - tx->GetDebit(ISMINE_STAKABLE);
```
aguycalled: ignoring ISMINE_SPENDABLE_STAKABLE would make this to fail for wallets holding both the spending and staking key (not an ideal scenario from a security perspective but its technically possible)

src/wallet/rpcwallet.cpp
```diff
@@ -3149,6 +3149,55 @@ struct StakePeriodRange_T {
...
+
+// Gets timestamp for first stake
+// Returns -1 (Zero) if has not staked yet
+int64_t GetFirstStakeTime()
+{
+    // Need a pointer for the tx
+    const CWalletTx* tx;
+
+    // scan the entire wallet transactions
+    for (auto it = pwalletMain->mapWallet.end(); it != pwalletMain->mapWallet.begin(); --it)
```
aguycalled: maps are ordered per key so theres no assurance the first stake you will hit is the oldest stake, i'd say the easiest would be to  iterate over wtxOrdered

src/wallet/rpcwallet.cpp
mxaddict: I'll see if I can add more scenarios to the test case

src/wallet/rpcwallet.cpp
mxaddict: ACK, I will replace this with wtxOrdered, thanks for the heads up.

mxaddict: I pushed a commit which addresses this.

src/wallet/rpcwallet.cpp
mxaddict: I think this scenario was already covered by the test cases

Please check this code in the test and let me know what you think:
```python
        # Import the 2 keys into a third wallet
        self.nodes[0].importprivkey(spending_address_private_key)
        self.nodes[0].importprivkey(staking_address_private_key)
```

Basically in the test cases, there are 3 modes, node 1 has spending key, node 2 has staking key and node 0 has both keys imported

aguycalled: yes you are right, my bad as `ISMINE_SPENDABLE_STAKABLE = ISMINE_SPENDABLE | ISMINE_STAKABLE`

mxaddict: I forget why, but in the 4.6.0 fix that I created for this

This would not get the correct amounts:
```cpp
tx->GetCredit(ISMINE_SPENDABLE_STAKABLE) - tx->GetDebit(ISMINE_SPENDABLE_STAKABLE);
```

Bu this would:
```cpp
tx->GetCredit(ISMINE_SPENDABLE) + tx->GetCredit(ISMINE_STAKABLE) - tx->GetDebit(ISMINE_SPENDABLE) - tx->GetDebit(ISMINE_STAKABLE);
```

While in theory they should be the same, but I don't remember exactly why.

I think there was code in `GetCredit` and `GetDebit` that was causing an issue when passing `ISMINE_SPENDABLE_STAKABLE`

aguycalled: `ISMINE_SPENDABLE_STAKABLE` is Spendable+Stakable

mxaddict: Anyways, I think the scenarios are covered in the test case

And I also added the time travel stuff that we discussed with @craigmacgregor during the 4.6.0 release which we skipped due to time constraints.

mxaddict:
Rebased the branch from master

proletesseract:
Just to confirm, all the `Last X Days Avg` values are programmed to give the daily average amount staked within period X or the daily average amount from the first stake till the end of X period if that length is shorter than X.

proletesseract:
Builds on Ubuntu 18.04. Tests pass. Tests make sense.

src/wallet/rpcwallet.cpp
```diff
@@ -3150,6 +3150,55 @@ struct StakePeriodRange_T {
...
+
+// Gets timestamp for first stake
+// Returns -1 (Zero) if has not staked yet
+int64_t GetFirstStakeTime()
+{
+    // Need a pointer for the tx
+    const CWalletTx* tx;
+
+    // scan the entire wallet transactions
+    for (auto it = pwalletMain->wtxOrdered.begin(); it != pwalletMain->wtxOrdered.end(); ++it)
```
proletesseract: If wtxOrdered is not changing length during this loop, you could cache the end iterator to save it being calculated each time the loop iterates. I'm not sure on the exact syntax here, in javascript you can do it directly into the loop declaration like this;

```
for(let i=0, l=my_array.length(); i < l; i++) {
    console.log(my_array[i]);
}
```
I don't think there's an inline way to do it in C++, so you'd just cache the iterator before you start the loop something like this;
```
std::vector<uint32_t>::const_iterator end = pwalletMain->wtxOrdered.end();
or (auto it = pwalletMain->wtxOrdered.begin(); it != end; ++it)
{
    //do stuff
}
```
It's probably not hugely relevant here since we're hoping the first item is the one we're looking for, but its a neat trick that can optimise loops especially when you're iterating through thousands of items like we can be doing when dealing with wallet transactions and especially long lived staking addresses. Keep it in mind if it's useful somewhere else :)

src/wallet/rpcwallet.cpp
proletesseract: I think this should be clearly documented somewhere so people don't accidentally assume the averages are simple averages over the entire X timeframe regardless of initial staking activity.

mxaddict:
@proletesseract Rebased on master and added cache for the wallet first stake value (Just in memory, so when the report is run more than once it will use the cached timestamp)

But this report will be slower for wallets with TX data but no stakes (Since it will have to loop all the TX every time in case it has staked since the last check)

proletesseract:
code looks good. need to resolve travis issues though.

mxaddict:
Travis-CI Passed, merging

PR #494, opened 2019-05-21
Use -rescan when -importmnemonic

aguycalled:
This PR solves a bug where the wallet did not rescan for transactions when importing a new seed.


rocket-pig:
awesome!

proletesseract:
works on ubuntu 18.04 for me too.

PR #500, opened 2019-05-29
TX_POOL missing in switch condition in the IsMine function

dantudor:
When TX_POOL was added I missed including it in the switch when considering IsMine()

It triggers a build warning which this PR resolves


aguycalled:
ACK

PR #502, opened 2019-05-30
Fixed some file permissions

mxaddict:
Fixed some file permissions


mxaddict:
Changes 755 to 644 permissions for images and all the `.md, .c, .cpp, .h, .yml` files that I could find

PR #507, opened 2019-06-05
Fixed a bug with the errorLogFile causing crashes

mxaddict:
@aguycalled I've fixed the issue you found.


aguycalled:
tested on OSX, it still crashes using devnet with an empty data folder
`./src/qt/navcoin-qt -datadir=/Users/alex/navcoin-core/data -printtoconsole -devnet -staking=0`

aguycalled:
tested OSX

PR #509, opened 2019-06-05
Updated some NULL -> nullptr

### LINKED ISSUE #508: NULL vs nullptr
I've noticed a lot of code that uses NULL and remembered that we use c++11 during compile

@aguycalled  @proletesseract Maybe we should replace all NULL with nullptr? Or is there a good reason to keep using NULL?

mxaddict:
closes #508


mxaddict:
Rebased on master to fix merge conflicts

mxaddict:
Rebased PR on latest master branch

mxaddict:
Merging into master, but we should think about adding a lint step to travis to check for these code changes in the future.

PR #536, opened 2019-06-19
Check node address in header spam protection

aguycalled:
Taken from QTUM fix https://github.com/qtumproject/qtum/pull/703


proletesseract:
compiles and runs on Ubuntu 18.04. Syncing works. Code changes make sense. I haven't tested spamming headers manually. Is it worth trying to write a unit test to cover this spam protection?

PR #537, opened 2019-06-20
Make some tests more reliable

mxaddict:
Updated a test to be more reliable (This test would randomly fail)


mxaddict:
Updated `qa/rpc-tests/cfund-proposal-state-accept.py` as well

PR #517, opened 2019-06-08
Move nBits check to ContextualCheckBlock

aguycalled:
This PR moves the nBits check to ContextualCheckBlock.

PR #544, opened 2019-06-29
Revert nBits patch

aguycalled:
This PR reverts https://github.com/navcoin/navcoin-core/pull/517 as it introduced a bug which prevented the wallet to sync from scratch and the DoS attack vector which it tried to fix is already covered by the Header Spam protection filter.


mxaddict:
This PR now syncs from block 0 with no errors from nBit check

Currently my test node it @ block ~150k

Will post back when fully synced

proletesseract:
building now, will report back soon.

mxaddict:
Tested with a full sync with all index flags enabled and completed successfully

PR #549, opened 2019-07-01
Updated __poll_chk implementation to be not so strict

mxaddict:
This fixes a crash for navcoin-qt when libs call poll() with bad data.

Updated the `__poll_chk` to behave less aggressive (More like `__explicit_bzero_chk`)

Calls `__chk_fail()` instead of an assert which would terminate the execution of the rest of the app

### DIFF
--- a/src/compat/glibc_compat.cpp
+++ b/src/compat/glibc_compat.cpp
@@ -6,7 +6,6 @@
 #include <config/navcoin-config.h>
 #endif
 
-#include <assert.h>
 #include <cstddef>
 #include <cstdint>
 #include <errno.h>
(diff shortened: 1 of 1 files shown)


mxaddict:
@marcus290 this is a fix for the issue you found.

mxaddict:
Basically I copied logic from glibc: https://sourceware.org/git/?p=glibc.git;a=blob;f=debug/poll_chk.c;h=34596de9777b3fe9f05e8a57840dd75076c51672;hb=HEAD

aguycalled:
Where do the original changes from https://github.com/navcoin/navcoin-core/commit/d3b56870eba55fbd2a0e323af6651f47d31e8b64#diff-626fe822220cf08b83b6459ec6c2b29b which introduced the bug come from?

mxaddict:
`glob_old`, `wrap_glob` and `__wrap_glob64` I coded based on the `__wrap_log2f` code that was already in the file.

`__poll_chk` I based on glibc implementation, but I used assert (I did not know `__chk_fail` existed) : https://sourceware.org/git/?p=glibc.git;a=blob;f=debug/poll_chk.c;h=34596de9777b3fe9f05e8a57840dd75076c51672;hb=HEAD

`__explicit_bzero_chk` I based on monero implementation: https://github.com/monero-project/monero/blob/master/src/common/compat/glibc_compat.cpp

`reallocarray` from : libunbound in our `./depends` dir

`getentropy` I think I coded myself.

mxaddict:
I updated my commit to also remove the header for assert

proletesseract:
building in ubuntu 18.04

PR #558, opened 2019-07-07
Added an additional LOCK for cs_main

mxaddict:
This fixes a bug I found in the RC build

Wallet would crash in certain scenarios due to a missing lock on `cs_main`


mxaddict:
Found another missing lock, would sometimes crash when setting votes as cs_main lock was not help when calling CFund::VoteProposal

PR #560, opened 2019-07-09
Set pindexDelete to nullptr on intialize

mxaddict:


mxaddict:
@chasingkirkjufell this PR is just to fix a warning on compile time.

PR #566, opened 2019-07-14
Updated includes syntax based on discussion on #503

### LINKED ISSUE #503: #include <> vs #include "" for headers
I noticed that upstream bitcoin-core has changes related to

```
#include <header.h>
// VS
#include "header.h"
```

Referenced here: https://github.com/bitcoin/bitcoin/pull/11651

Do you guys think we should do the same?

It makes since to have all header includes use `<>` as it will make the files included relative to root of the project instead of relative to the file including it.

mxaddict:
closes #503


mxaddict:
Rebased on latest master branch

PR #591, opened 2019-08-24
Patch UpdatePrice if curl don't get data on WIN

R-Marino:
In some cases the UpdatePrice function does not update the data
because curl cannot establish a secure connection.

With this fix, if curl has reported error code 60, attempts a new
connection without peer verification, like DownloadBlockchain() do.

If the remedy works, this shortcut will be used for the rest of the
session to recover the prices.
Otherwise it will not try the remedy again.

### DIFF
--- a/src/qt/navcoingui.cpp
+++ b/src/qt/navcoingui.cpp
@@ -1653,7 +1653,7 @@ void NavCoinGUI::updateWeight()
 
 void NavCoinGUI::updatePrice()
 {
-    info("Updating prices");
+    static int nBypassSSLVerify = 0; // not using it
 
     std::thread pThread{[this]{
         try {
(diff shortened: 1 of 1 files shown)


aguycalled:
If the SSL certificate is not checked, we can't trust that the source of the data is what we expect to be. Showing an incorrect fiat value can have consequences if someone takes it as valid and uses it as reference for manual calculations.
Maybe it would be better to just show `N/A` (not available) when the price can't be pulled. Thoughts?

mxaddict:
I think I'm with aguycalled with this one, maybe we need to build windows curl agains windows ssl instead of openssl.

I've seen this fix similar issues in the past.

mxaddict:
I'll see if I can create a build that uses windows ssl lib instead of openssl

R-Marino:
Exactly what I tried to do this morning.
I changed the file curl.mk (line 16) with:
  $(package)_config_opts_mingw32=--with-random=/dev/urandom --with-winssl --without-ssl

The curl.exe file generated with winssl (Schannel) in this way seems to work.

NavcoinQt (with the current commit) does not trigger the "bypass", and not shows errors during updatePrice(), then "seems" to work.

Do I drop  this commit?

R-Marino:
> If the SSL certificate is not checked, we can't trust that the source of the data is what we expect to be.

I think the same thing can be said for loading the bootstrap file, even with worse consequences.


> "Showing an incorrect fiat value can have consequences if someone takes it as valid and uses it as reference for manual calculations.
> Maybe it would be better to just show N/A (not available) when the price can't be pulled. Thoughts?"

The proposed fix was a fallback, used only if the direct connection had not worked. A last chance.

But you're right, if due to a failure to update the price data is obsolete, it should be reported in some way.
In a very volatile market, a few minutes delay can lead to big differences.

mxaddict:
@R-Marino if you were able to compile with the curl.mk changes you mentioned (using -winssl).

I think you should add those changes and revert this fallback logic (As compiling with windows ssl is a more elegant solution)

Maybe a rebase or a new PR will be better for a cleaner git history.

mxaddict:
@aguycalled  @R-Marino fix for this is now in #601 I've so far tested on Windows 10 64bit and will test on OSX once I get hands on a mac

mxaddict:
Closing this as it's confirmed fixed for both windows and osx in issue #603

PR #592, opened 2019-08-27
Fixed permission issue that was causing travis to fail

mxaddict:


mxaddict:
@R-Marino I've created this PR with permission fix, but seems that the travis build still fails, I'm not sure why cause running the testsuit locally passes fine.

I literally can't replicate the issue on travis

mxaddict:
@aguycalled  @R-Marino I've found the cause for the issue.

Seems that the local NTP server is not allowing more connections (I'm assuming this is because of the large number of nodes that the test needs.

I'm adding a code change to allow for ntpminmeasure=-1 (Which will be like the 0 value, but will suppress the errors/warnings that come with the ntpminmeasures=0 value.

aguycalled:
That’s fantastic. Adding `ntpminmeasures=-1` to the default navcoin.conf file for tests would solve this and probably other test issues.

mxaddict:
> That’s fantastic. Adding `ntpminmeasures=-1` to the default navcoin.conf file for tests would solve this and probably other test issues.
> […](#)
> On 29 Aug 2019, at 16:43, Barry Deeney ***@***.***> wrote: @aguycalled @R-Marino I've found the cause for the issue. Seems that the local NTP server is not allowing more connections (I'm assuming this is because of the large number of nodes that the test needs. I'm adding a code change to allow for ntpminmeasure=-1 (Which will be like the 0 value, but will suppress the errors/warnings that come with the ntpminmeasures=0 value. — You are receiving this because you were mentioned. Reply to this email directly, view it on GitHub, or mute the thread.

My thoughts exactly

aguycalled:
i've tested locally `-ntpminmeasures=-1`, it still tries to ntp sync. the right would be `-ntpminmeasures=0`

mxaddict:
@aguycalled looks like these changes are working.

aguycalled:
travis passed

proletesseract:
change makes sense, havent locally compiled but travis passes

Issue #584, opened 2019-08-09
Implement a GUI for re-directing stakes


chasingkirkjufell:
Having the ease of use for re-directing stakes could encourage more donations for development and DAO.

chasingkirkjufell:
Resolved with #605

PR #593, opened 2019-08-28
Adds text field in settings UI to send stakes to a different address

aguycalled:
Solves https://github.com/navcoin/navcoin-core/issues/584


mxaddict:
utACK

PR #604, opened 2019-09-16
Max reorganization depth

aguycalled:
This PR sets a max reorganisation depth of 500 blocks.

Tested locally with `art-of-bug` details.

```
2019-09-16 18:49:17 Misbehaving: 127.0.0.1:12312 (0 -> 1)
2019-09-16 18:49:17 ERROR: invalid header received
2019-09-16 18:49:17 ProcessMessages(headers, 155803 bytes) FAILED peer=1
2019-09-16 18:49:17 Misbehaving: 127.0.0.1:12312 (1 -> 2)
2019-09-16 18:49:17 ERROR: invalid header received
2019-09-16 18:49:17 ProcessMessages(headers, 155803 bytes) FAILED peer=1
2019-09-16 18:49:17 Misbehaving: 127.0.0.1:12312 (2 -> 3)
```

### DIFF
--- a/qa/pull-tester/rpc-tests.py
+++ b/qa/pull-tester/rpc-tests.py
@@ -187,6 +187,7 @@ testScripts = [
     'sendtoaddress.py',
     'stakeimmaturebalance.py',
     'rpc-help.py',
+    'max-reorg-depth.py',
 ]
 #if ENABLE_ZMQ:
 #    testScripts.append('zmq_test.py')
--- a/qa/rpc-tests/max-reorg-depth.py
+++ b/qa/rpc-tests/max-reorg-depth.py
@@ -42,14 +42,9 @@ class MaxReorgDepth(NavCoinTestFramework):
         blocks0 = self.nodes[0].getblockcount()
         blocks1 = self.nodes[1].getblockcount()
 
-        cur_time = int(self.nodes[0].getblock(self.nodes[0].getbestblockhash())["time"])
-
         filter_size = 500
 
         for i in range(filter_size+1):
-          cur_time += 30
-          self.nodes[0].setmocktime(cur_time)
-          self.nodes[1].setmocktime(cur_time)
           self.nodes[0].generate(1)
           self.nodes[1].generate(1)
 
--- a/qa/rpc-tests/max-reorg-depth.py
+++ b/qa/rpc-tests/max-reorg-depth.py
@@ -25,6 +25,7 @@ class MaxReorgDepth(NavCoinTestFramework):
 
     def run_test(self):
 
+        self.nodes[0].staking(False)
         self.nodes[1].staking(False)
         slow_gen(self.nodes[0], 1000)
         self.sync_all()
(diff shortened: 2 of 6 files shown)


mxaddict:
utACK

mxaddict:
Changes, look good, will test using art of bug details in a few minutes.

proletesseract:
building on OSX

proletesseract:
Code makes sense, build works. looking into possibility of adding RPC tests.

R-Marino:
Hi! I haven't tested it, but if I can express an opinion, at first
sight I don't like it, for various reasons.

1. Efficiency.
If I understand the code well, for each header received out of
sequence, the last 500 blocks are scanned backwards one by one.
If only 500 spam headers are sent, this is equivalent to performing
250,000 iterations.  These iterations even if performed in memory
lead to a useless slowdown.
I don't dare to think about what could happen to a node that has
to synchronize itself after being stopped for a few weeks.
     
I think you can easily mitigate this aspect by totally removing
the AutoSelectSyncCheckpoint routine and using GetAncestor instead,
or even more directly:
`   chainActive[max(chainActive.Height() - nCheckpointSpan, 0)]`
or something similar. This do the same, but requires only 1 access
to a vector instead of 500 sequential iterations on pointers.
     
2. Efficacy of the fix
Increasing the MaxSize size of the filter to 500, keeping the ban limit
to 3 * MaxSize, means that a spammer can send up to 1500 spam before
being banned.   In some ways the fix worsens the previous situation.
I think You can mitigate the fix by adjusting MaxSize or organizing
differently the limits of the ban and the size of the filter.
     
3. Persistence of spam and waste of resources.
With this system all the spam received is still recorded on disk, forever.
     
Perhaps it could be more useful to keep the block and the index of the
header only in memory and write them on disk only when the block has 
actually been verified.
If, on the other hand, the block just received is invalid for serious
reasons, then the data of the subsequent blocks should be taken out
of memory and their hash placed in a temporary blacklist that uses much
less memory than the CBlockIndex.
     
Of an invalid spam sequence, only the first invalid block would be
permanently recorded, and even the memory would be largely retrieved.
Obviously this is a more complex solution and requires more study.

4. Resilience Vs Autocheckpoint
Personally, I am not very keen on the autocheckpoint system.
     
In the past we have seen single honest nodes that for some reason
generated many blocks, in solitary forks.
     
With an autocheckpoint system, the forked node could stubbornly 
consider the blocks created by the rest of the network as spam,
since the correct blocks are not compatible with its sequence of
old (erroneous) blocks.
     
So the forked node will start banning the honest nodes one by one
and then it will never be able to synchronize properly, even after
multiple restarts.
     
I think it might be useful to have a different system or an option
that in such disperate cases can temporarily add a recent manual
checkpoint, for example taken from one of the block explorers and 
that can help bypass the autocheckpoint and facilitate automatic
synchronization.
     
There are probably other more complex systems that would avoid all
these problems with header spam, but perhaps they would require
more Byzantinisms.

aguycalled:
@R-Marino I strongly appreciate your input. I agree with point 1 and I've incorporated your suggestion in https://github.com/navcoin/navcoin-core/pull/604/commits/9d657546f8bbd8e6e3af04318f0e53899178a3e8.
Point 2: Could you be more specific about it?
If we reduce MaxSize we would also need to reduce the maximum reorg depth accordingly so we are protected against header spam in the conditions described by art-of-bug. Reducing the max reorg depth would reduce the capability of a node to reorganise in honest conditions, which is not desirable. But I might be missing something here.
As I pointed in Discord when this was disclosed I'm also not keen at all on having an auto-checkpointing system. It's important to note this is a temporary fix. From my perspective the redesign must go even deeper than point 3, which I have to say it is a well though solution but not enough for our problematic. We need to address the fact that with the current design, there's nothing one can really validate from a block header (besides the deployment signal and block's timestamp) without having an accurate up-to-tip view of the utxo set using the data coming from the full block. The headers-first-sync approach makes no sense in the PoS design we use, because one can't correctly validate which one is the right chain in a headers-only view. That's what needs a revault and where I would love to see the dev efforts focused sooner than later.

aguycalled:
By the way, @R-Marino have you received our email? We would love to confirm in which address you would like to receive the pending bounty payments.

mxaddict:
rebuilt and ran automated test, seems to work reliably

R-Marino:
If I remember correctly, a reorganization takes place only when a
fork has more blocks and or has a greater chain work compared to
the other fork.

At first glance, it seems to me that in the new test script the two
nodes have the same number of blocks before the sync.
Thus, the re-org simply cannot happen.

Still on the subject of test scripts, it is possible to arrange the 
various nodes on different ip, for example 127.0.0.2, 127.0.0.3, 
and so on?   If yes, how?
This would probably help to control the ban mechanism.

@aguycalled
Yes Alex, I am very late with emails and other activities, but I received it,
thanks very much.

aguycalled:
@R-Marino you can use `-bind=ip` to bind a node to a specific ip

proletesseract:
What's the plan with this one? Back to the drawing board?

aguycalled:
yep, closing

PR #605, opened 2019-09-20
Staking reward setup GUI

aguycalled:
This PR adds a new GUI to set up different ways to split the staking rewards. Setup can be done generally for every address or individually per staking address.

Works for both hot and cold staking methods.

Includes the following hardcoded addresses:

```
{"NN5QSSMAdtRU35BffLZUw9vChnhHKKMeuL", "Alex aguycalled - Core Dev"},
{"NRXfZ1egFxMSUsc4Ufpi4Lm7DdXStYmeBG", "mxaddict - Core Dev"},
{"3HnzbJ4TR9", "Community Fund"},
{"XVPMwBdNU9ou3a3TnwaVgAgEecbdsEVZHbVmeY4TMAHbY6BdtY8xW6m1Q1rkb", "Core Dev Bounty"},
{"XAznGHuQ35hvgSGsVWi5Nu2Y6n3rT4cycE3yfZWCfnNjycCGdGAEnta2G24Mi", "Web Dev Bounty"}
```

Please comment in this pull request to add other community member addresses.

image
image

### DIFF
--- a/src/qt/overviewpage.h
+++ b/src/qt/overviewpage.h
@@ -62,7 +62,7 @@ public Q_SLOTS:
     void updateStakeReportNow();
     void updateStakeReportbalanceChanged(CAmount, CAmount, CAmount, CAmount, CAmount, CAmount, CAmount);
     void setVotingStatus(QString text);
-    void on_pushButton_clicked();
+    void on_showStakingSetup_clicked();
 
 Q_SIGNALS:
     void transactionClicked(const QModelIndex &index);
--- a/src/qt/overviewpage.h
+++ b/src/qt/overviewpage.h
@@ -6,6 +6,7 @@
 #define NAVCOIN_QT_OVERVIEWPAGE_H
 
 #include "amount.h"
+#include "splitrewards.h"
 
 #include <QWidget>
 #include <QPushButton>
@@ -61,11 +62,11 @@ public Q_SLOTS:
     void updateStakeReportNow();
     void updateStakeReportbalanceChanged(CAmount, CAmount, CAmount, CAmount, CAmount, CAmount, CAmount);
     void setVotingStatus(QString text);
-
+    void on_pushButton_clicked();
 
 Q_SIGNALS:
     void transactionClicked(const QModelIndex &index);
-    void outOfSyncWarningClicked();
+    void outOfSyncWarningClicked();    
 
 private:
     Ui::OverviewPage *ui;
--- a/src/qt/splitrewards.cpp
+++ b/src/qt/splitrewards.cpp
@@ -79,7 +79,11 @@ SplitRewardsDialog::SplitRewardsDialog(QWidget *parent) :
     {
         CTxDestination dest;
         if (ExtractDestination(out.tx->vout[out.i].scriptPubKey, dest)){
-            QString strAddress = QString::fromStdString(CNavCoinAddress(dest).ToString());
+            CNavCoinAddress address(dest);
+            if (address.IsColdStakingAddress(Params()))
+                if (!address.GetStakingAddress(address))
+                    continue;
+            QString strAddress = QString::fromStdString(address.ToString());
             if (mapAddressBalance.count(strAddress) == 0)
                 mapAddressBalance[strAddress] = 0;
             mapAddressBalance[strAddress] += out.tx->vout[out.i].nValue;
--- a/src/qt/splitrewards.cpp
+++ b/src/qt/splitrewards.cpp
@@ -26,9 +26,9 @@ SplitRewardsDialog::SplitRewardsDialog(QWidget *parent) :
     topBoxLayout->setContentsMargins(QMargins());
     topBox->setLayout(topBoxLayout);
 
-    topBoxLayout->addWidget(new QLabel(tr("<b>Staking Rewards</b>")));
+    topBoxLayout->addWidget(new QLabel(tr("Configure where your staking rewards are forwarded.")));
     topBoxLayout->addStretch(1);
-    topBoxLayout->addWidget(new QLabel(tr("Setup for Staking Address:")));
+    topBoxLayout->addWidget(new QLabel(tr("Setup for staking address:")));
     topBoxLayout->addWidget(comboAddress);
 
 
@@ -52,7 +52,7 @@ SplitRewardsDialog::SplitRewardsDialog(QWidget *parent) :
     QPushButton* quitBtn = new QPushButton(tr("Cancel"));
     connect(quitBtn, SIGNAL(clicked()), this, SLOT(onQuit()));
 
-    connect(comboAddress, QOverload<int>::of(&QComboBox::currentIndexChanged),
+    connect(comboAddress, QOverload<int>::of(&QComboBox::activated),
         [=](int index){ showFor(comboAddress->itemData(index).toString()); });
 
     bottomBoxLayout->addWidget(addBtn);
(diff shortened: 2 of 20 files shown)


chasingkirkjufell:
built and tested locally on ubuntu 18.04. it worked as expected. made a couple suggestions on handling the default settings.
the receiving stakes will display 2 NAV no mater how ever much the % is if the staking address and the receiving addresses both belong to the same wallet.dat. not likely to be encountered and the actual amount is accurate despite showing 2 NAV.

chasingkirkjufell:
image

the new drop down is empty and nothing happens after clicking.

aguycalled:
should be fixed now @chasingkirkjufell

chasingkirkjufell:
in cold-staking mode, staking node stake redirection did not work when choosing the cold-staking address from the drop down list and then point to an address that belongs to staking node. it only worked when the redirection address was set under "default" rather than "cold-staking address".

aguycalled:
this happens because the wallet looks for the configuration for the staking address of the cold staking address which owns the coins that are staking. i've updated the gui so it only shows the staking address instead of the long cold staking one. does it sound like a good solution?

chasingkirkjufell:
should be fine if the amount it's cold-staking is also shown so users don't have to figure out which address they used. the questions though will be what will it show if somehow someone send coins to the staking address, will it show the cold-staking amount or the amount in the address? will test it tomorrow.

aguycalled:
it will show both

chasingkirkjufell:
when receiving re-directed hot-staking or cold-staking, the transaction history shows receiving 0 NAV instead of the redirected amount on the re-directed address. 
image

image


coin control shows the correct amount.

image

aguycalled:
how is it shown in `listtransactions`?

chasingkirkjufell:
0

image

aguycalled:
should be fixed now

chasingkirkjufell:
setting window width increases with the amount of addresses set and cannot be decreased.

image

image

http://giphygifs.s3.amazonaws.com/media/njYrp176NQsHS/giphy.gif :P

aguycalled:
fixed
image

chasingkirkjufell:
users are able to set 0% contribution to an address, which is fine, but the receiving address will actually see it. could it be used to bloat the blockchain if someone just set infinite numbers of addresses getting 0%?

image

image

chasingkirkjufell:
when setting both default redirection address and individual address redirection address, redirection will stop after setting individual address and then removing it while it shouldn't since the default should still redirect it to the set address. 

a uncommon situation but when cold-staking, the spending wallet redirection drop down will show the staking address as well as the spending address tied to it as separate addresses. uncommon since if someone is cold-staking, they won't need to use redirection on the spending wallet.

chasingkirkjufell:
let me know when it's ready to be  tested :)

aguycalled:
ready

src/qt/splitrewards.cpp
```diff
@@ -175,6 +175,17 @@ void SplitRewardsDialog::setModel(WalletModel *model)
 
 void SplitRewardsDialog::onAdd()
 {
+    if (availableAmount <= 0)
+    {
+        QMessageBox msgBox(this);
+        msgBox.setText(tr("You are already contributing the 100% of your staking rewards."));
```
chasingkirkjufell: i think the word "the" is redundant?

chasingkirkjufell:
one can see the cold-staking staking address (not the long cold-staking address) in the drop down of the setting window while in spending wallet. it may not be important but still worth hiding so if a user accidentally loaded the spending wallet, the drop down list can be empty since the user shouldn't be staking with a spending wallet. can the user even stake with spending wallet if already have all coins in a cold-staking address?

aguycalled:
no, the spending address can't stake

chasingkirkjufell:
whew, this one was a tough one to figure out. 
first, 
cold-staking redirected to another cold-staking address doesn't seem to work

second,
when a node is both cold-staking and hot-staking and the hot-staking reward redirection is towards a cols-staking address (won't happen if redirected to normal address), hot-staking addresses will show incorrect amount when mined a block, the contribution seems to be counted twice when a hot-staking address mines a block. say 35% contributed to another address, when a block is mined, it'll show 0.7 NAV redirected and 0.6 NAV incoming. it'll even be -2 NAV is 100% redirected.

chasingkirkjufell:
I might have over-complicated the second point, the node might not need to be both cold-staking and hot staking. it might be just that normal address redirected to cold-staking address will show incorrect amount. will test it tomorrow.

aguycalled:
```
first,
cold-staking redirected to another cold-staking address doesn't seem to work
```
did you setup for the cold staking address or the staking address from the cold staking address?

chasingkirkjufell:
both, a cold-staking node doesn't redirect funds to another long cold-staking address. It doesn't matter if the setting is in default or in the staking address of the cold-staking address, neither work.

chasingkirkjufell:
wait, i might have made a mistake

chasingkirkjufell:
my bad, cold-staking redirecting to another cold-staking address works just fine. i opened the wrong wallet and redirected to myself that was why i couldn't see any contribution difference. and a hot-staking node can redirect stakes to a cold-staking address normally. so the irregularity is only when a node is both hot-staking and cold-staking.

chasingkirkjufell:
okay, narrowed down. if a node is cold-staking and hot staking, hot staking stake redirection works normally unless the redirected address is the long cold-staking address which the wallet owns the staking address, then the incorrect amount will be shown. the correct amount is shown if hot-stakes redirected to just the staking address of the cold-staking address. only doesn't work when redirected to the long cold-staking address the wallet is staking with.

image

aguycalled:
could you share a screenshot of your rewards setup?

chasingkirkjufell:
image

image

image

image

aguycalled:
For me it shows correct.
image
mgo.. and mvX... are respectively the staking and spending addresses of the long cold staking where the coins sit
image

aguycalled:
image
mgo and mxV are the staking and spending addresses of the long cold staking addresses
image
maybe i'm doing something different?

chasingkirkjufell:
Yeah the spending address in my situation belongs to another wallet. The problem comes when a normal address of the staking node and belongs to the staking node wallet finds a block and redirects it to a long cold staking address of which the staking node only owns the staking address and not the spending address

chasingkirkjufell:
Kind of proud that I could find that problem tbh :P

mxaddict:
That's a pretty niche issue. But i guess it's a valid use case.

aguycalled:
@chasingkirkjufell  i wanted to send you some navs from the dev bounty for helping testing if you could share a nav address :wink:

chasingkirkjufell:
sure. really appreciate it. :)
NL9jPW75P4kNMdvQmFBpYVkZ3sABQtXfYY

mxaddict:
❤️❤️❤️

aguycalled:
i think i got it fixed now

chasingkirkjufell:
tested and worked well. all redirection possibilities worked as expected. the only imperfection is what i mentioned before but it doesn't affect functionality. good job! 

https://github.com/navcoin/navcoin-core/pull/605#issuecomment-535217176

mxaddict:
@aguycalled can you add my address to this list? :)

NRXfZ1egFxMSUsc4Ufpi4Lm7DdXStYmeBG

src/qt/splitrewards.h
```diff
@@ -24,6 +24,7 @@
 #include <QVBoxLayout>
 
 static QMap<QString, QString> teamAddresses = {{"NN5QSSMAdtRU35BffLZUw9vChnhHKKMeuL", "Alex aguycalled - Core Dev"},
+                                               {"NRXfZ1egFxMSUsc4Ufpi4Lm7DdXStYmeBG", "mxaddict - Core Dev"},
```
mxaddict: :heart:

aguycalled:
Paid 2,800NAV to NL9jPW75P4kNMdvQmFBpYVkZ3sABQtXfYY
300 (minor ux issue) + 600 (major ux issue) + 600 (major ux issue) + 600 (major ux issue) + 100 (visual glitch) + 300 (minor ux issue) + 300 (minor ux issue)
339db52e68ca2937bc36fe38addca5d6e15333cbaf54de4045d76400d0e05aed

chasingkirkjufell:
thanks a lot!! :D

proletesseract:
Compiles & Runs on OSX 10.14.5. Seems like there's been some pretty rigorous testing on this one so far. My only real feedback is adding a button to the staking section on the overview page which also launches the staking setup window to drive people to consider using it more than just hidden away in the menu. I've committed the change adding the button. If you don't think we should have the button there we can roll it back.

I've also added a donation address for me :)

Screen Shot 2019-10-06 at 5 08 24 PM

src/qt/overviewpage.h
```diff
@@ -61,11 +62,11 @@ public Q_SLOTS:
     void updateStakeReportNow();
     void updateStakeReportbalanceChanged(CAmount, CAmount, CAmount, CAmount, CAmount, CAmount, CAmount);
     void setVotingStatus(QString text);
-
+    void on_pushButton_clicked();
```
aguycalled: @proletesseract could we maybe use another name for this method? it reads very generic and does not refer to its real meaning

proletesseract: Sure, that was just what was autogenerated by QT Creator. I'll commit the change when i get a chance.

PR #608, opened 2019-10-06
Fix CFund DB read after nullified entry

aguycalled:
This PR fixes a bug where a cfund db entry previously nullified is still read as existent preventing future writes.

This can happen in a scenario where a proposal is included in a block which is invalidated (because of a reorganization) for later being included in a new block part of the new chain.

The bug is explained because FetchProposal was not considering previously nullified entries as non existent preventing AddProposal from correctly adding the new one.

Includes test.


mxaddict:
utACK, building now

mxaddict:
unit test makes sense.

mxaddict:
Test passed locally, tested on Ubuntu 18.04

qa/rpc-tests/cfund-proposal-reorg.py
```diff
@@ -0,0 +1,51 @@
...
+        self.is_network_split = split
+
+    def run_test(self):
+
+        activate_cfund(self.nodes[0])
+        sync_blocks(self.nodes)
+
+        rawproposal = self.nodes[0].createproposal(self.nodes[0].getnewaddress(), 10, 36000, "test", 50, True)
+
+        # disconnect the nodes and generate the payout on each node
```
proletesseract: does this generate the payment request on each node, or just the proposal? Is it worth adding a test to cover the payment request as well?

qa/rpc-tests/cfund-proposal-reorg.py
```diff
@@ -0,0 +1,51 @@
...
+        hash = self.nodes[0].sendrawtransaction(rawproposal)
+        self.nodes[1].sendrawtransaction(rawproposal)
+
+        self.nodes[0].generate(1)
+        self.nodes[1].generate(2)
+
+        connect_nodes_bi(self.nodes, 0, 1)
+        sync_blocks(self.nodes)
+
+        assert(self.nodes[0].getproposal(hash) == self.nodes[0].getproposal(hash))
```
proletesseract: i think this is supposed to compare node 0 to node 1, not just node 0 to node 0 ?

I'm trying to add checking the payment request reorg to this test, so i can fix it in my commit.

proletesseract:
@aguycalled easier to comment here than discord as we have many parallel discussions there. I am happy the fix for re-org of a proposal is working correct. I corrected the test in commit https://github.com/navcoin/navcoin-core/pull/608/commits/b502b812cc6c297ae00ba92ca811fafb77cfd208. I tried to write a test to check the payment request reorg operates as intended but was unsuccessful. If you think we should roll that up into this PR i can commit my test (which gives the unexpected result) into this branch for your review and reverse my approval.

mxaddict:
utACK on the changes, building now

PR #625, opened 2019-11-12
CFundDB state hash check

aguycalled:
This pull request proposes a new method to prematurely detect state consistency between nodes before they cause consensus disagreements.

Nodes with this feature will add to their generated blocks a commitment to the state hash `C=H(P_0||...||P_i||Pr_0||...||Pr_j)`, where `H` is the SHA256 hash function, `P` is a vector of size `i` with the Proposals present in the CFundDB and`Pr` is a vector of size `j` with the Payment Requests, both in lexicographic order.

The commitment will be added as an output with value 0 in an arbitrary position of the outputs' vector from the coinbase transaction in the form:
```
OP_RETURN OP_CFUND_HASH C
```
Where `OP_CFUND_HASH=0xcc` and `C` is the aforementioned commitment.

Nodes will compare the broadcasted state hash with their own, and will show a warning to the user `It looks like your state database might be corrupted. Please close the wallet and reindex.` when more than `-cfunddbstatehashthreshold` of the last `-cfunddbstatehashcount` blocks diverge from their own value. Headless nodes will be able to look for this message in the `errors` key of the `getinfo` rpc command.

For this method to be effective it is required that a majority of stakers upgrade to a version supporting the state hash broadcast feature. The values of `-cfunddbstatehashthreshold` and `-cfunddbstatehashcount` can be adjusted accordingly on a per-node basis to ensure state inconsistencies are detected in a determined time frame depending on the concrete network conditions.

By default, the wallet will use the values`-cfunddbstatehashthreshold=21` and `-cfunddbstatehashcount=60`, which allow to detect state inconsistencies in a maximum of 30 minutes assuming:
- 30 seconds block time
- 70% of the newly generated blocks broadcasting the state hash 
- the network fraction with a diverged state holds more than 1/2 of the staking weight

`-cfunddbstatehashcount` is calculated as `desired time frame` divided by `block time`
```
30 min / 0,5 min = 60
```

`-cfunddbstatehashthreshold` is calculated as `-cfunddbstatehashcount` * `percentage of blocks broadcasting state hash` * `staking weight threshold`
```
60 * 0,7 * 0,5 = 21
```

Includes a functional test `qa/rpc-tests/cfunddb_statehash.py`, adds three new arguments:

`-fakecfunddbstatehash` only valid in devnet and intended for testing, sets the broadcasted state hash to a fixed fake value
`-cfunddbstatehashthreshold` sets the threshold for triggering cfunddb state warnings
`-cfunddbstatehashcount` sets how many blocks are counted for cfunddb state warnings

and a new rpc command:

`getcfunddbstatehash` returns the current value of the state hash.


mxaddict:
@aguycalled the new test in this PR seems to fail in travis, but it looks like it passed.

Might be because of the stderr output from the daemon?

aguycalled:
yes that might be the cause. any idea how to make travis ignore the stderr output?

mxaddict:
I think if we start the daemon with >&2 after the command it should redirect the stderr.

aguycalled:
i wonder if the stderr output is checked for detecting failures in other tests

mxaddict:
I mean 2>&1

mxaddict:
We could just pass it for this specific test.

mxaddict:
I just got home, I'll see if I can redirect stderr just for this test.

R-Marino:
I only gave a quick read at the commit, so forgive any
inaccuracies in these first hot impressions.

If you feel the need to put in place such strict control on
the uniformity of the status of the proposals, this would
suggest that you believe the pre-existing problem has not
been overcome.
If I am not mistaken, the status of the proposals is stored
in the same database as the transaction not yet spent
(then chainstate).

This chainstate is much more relevant than the proposals,
do you think that it is also at risk of inconsistencies
and misalignments?



From a technical point of view, if I understand correctly, 
the hash that represents the status of all the proposals & 
request is calculated again:
 A) In the verification phase, for every block that is
    about to be connected to the chain, we assume about
    every 30 seconds.
    
 B) During the preparation of a new stake block, 
    approximately every 16 seconds.
    
However, the new proposals themselves are not generated
continuously, so its hash remains the same for days.

It would not seem more efficient, recalculate the hash
only when the proposals set change, and hashing only the
proposals still active?

.

Still on the subject of efficiency, I understand that the
nodes would register the hash in each block they generate,
even if it is the same as in the previous block.
In order to limit the waste of bytes, if the calculated hash
is still the same, don't you think that it would be more
useful to indicate a short code, for example a simple 0, 
instead of wasting 64 more bytes for the same hash?
(The blockchain is forever. Better to be as lean as possible)

.

Another argument, hypothesis of abuse for FUD.
A group of dishonest nodes could create valid blocks, but
with incorrect hashes for the sole purpose of making the
other honest nodes believe that something has gone wrong inside
their nodes and thus induce them to disconnect for a long time,
in an attempt to redo an unnecessary sync.
The dishonest nodes do not suffer any damage or penalty of
any kind in doing this.

.

Lastly, if one or more honest nodes create their fork for
some unforeseen reason, as has happened other times in the
past, they would not try to connect the blocks of the other
fork, nor compare their hashes.
So the forked nodes may not receive any notification of
misalignment, even if they are the ones most in need to know.

Xa5r:
"don't you think that it would be more
useful to indicate a short code, for example a simple 0,
instead of wasting 64 more bytes for the same hash?
(The blockchain is forever. Better to be as lean as possible)"

Yes I agree, efficiency should always be important in blockchain

aguycalled:
The chainstate implementation is untouched from the original codebase, so there's no risk of misfunctioning there. It's in the particularities of the management of the cfund entries on the view caches where the issues appeared. Once can see the differences in all the ::*Proposal* and ::*PaymentRequest* methods and additionally in ::BatchWrite() in CCoinsViewCache:: and CCoinsViewDB::.

The hash includes the serialized version of the proposals and payment requests, so it does not need to be recalculated only when new entries are added but every time there is a modification. In terms of efficiency, my impression is that it is more efficient to recalculate it only on validation or creation of a new block, but I don't have the concrete numbers.

Re: your suggestion of commiting to a 0 if the value does not change, it would be at the cost of having to read the full previous block (or a sequence of them until we find the last commited value) to extract. Another alternative would be to store it attached to the blockindex, but then we would be spending memory resources. Not completely sure if it's a better approach considering the reduced amount of data (32 bytes).

Forks always occur way later than when data base inconsistencies occur. For instance, proposal inconsistencies only cause forks when a payment request is created, and payment request inconsisntencies only cause forks when a payout for that payment request is included in a block. The goal of this pull request is to introduce a method to detect inconsistencies during a limited time when the community fund is going to be disabled, to help depurating real world scenarios and specially in a testing context. The ultimate goal is to have a depurated version of the current CFundDB which does not need this validation.

R-Marino:
> The chainstate implementation is untouched from the original codebase, so there's no risk of misfunctioning there.

Very well. This was something that needed to be reaffirmed.

> The hash includes the serialized version of the proposals and payment requests, so it does not need to be recalculated only when new entries are added but every time there is a modification.

If the serialisation also includes the number of votes accumulated, then you are right.
However, I do not see the reason for the inclusion of the status of all the proposals, even those that have long expired or can no longer change the status since many cycles


> In terms of efficiency, my impression is that it is more efficient to recalculate it only on validation or creation of a new block, but I don't have the concrete numbers.

Once a block has been received, connected and counted, the status hash must remain stable, at least until the bestblock change.
No need to do many calculations.

 
> Re: your suggestion of commiting to a 0 if the value does not change, it would be at the cost of having to read the full previous block from disk (or a sequence of them until we find the last commited value) to extract it, which is slow. Another alternative would be to store it attached to the blockindex, but then we would be spending memory resources. Not completely sure if it's a better approach considering the reduced amount of data (32 bytes).

Yes 32 bytes, not 64, sorry.
No thanks, the blockindex is already too big now, and above all this hash data does not need to be remembered forever.
At the bottom we are only interested in the status of the current round, so just over 20K blocks.

If we want to reduce more, it would probably have been sufficient to issue the hash also at regular intervals (ie 250 block or 10 Min) and we could reduce the use of memory to a cache limited to a few tens of hashes.
If, on the other hand, the hash also includes the information on the vote count and most of the blocks include votes, then the question loses its importance.

>The goal of this pull request is to introduce a method to prematurely (before a fork could be caused) detect inconsistencies during a limited time when the community fund is going to be disabled, to help depurating real world scenarios and specially in a testing context.

Disabled? What do you mean? 
Have I lost some key announcements?

navbuilder:
Your Message to Comment

navbuilder:
A new build of 31992f26ec92ea9c0421738ad642b592e9285f95 has completed succesfully!
Binaries available at https://build.nav.community/binaries/dao-state-hash

mxaddict:
@aguycalled I don't fully grasp how much more space would be occupied on the blockchain by adding this change, @R-Marino seems a bit concerned about it, do you have a quick math sample of the amount of data that would be added by this change?

aguycalled:
Closing this, focusing on tesnet stress testing.

PR #643, opened 2019-12-05
CPaymentRequest fields incorrect in diff

dantudor:
I haven't reviewed what is using the diff method for CPaymentRequest. 
I did notice that some of the conditions were incorrect by either the expression, the output text or format specifier


navbuilder:
A new build of d01e08eaf3706e0deb4628015ce2886cb363b340 has completed succesfully!
Binaries available at https://build.nav.community/binaries/patch-1

PR #644, opened 2019-12-06
Adds proposalHash to the RPC getpaymentrequest

dantudor:
The proposal hash has been added to the result of RPC getpaymentrequest.

A root flag has been added to not add the proposalHash when the payment request is not the root JSON node (when it is iterated as a child of RPC getproposal)


aguycalled:
ACK untested but approving as travis passed and changes are simple enough

navbuilder:
A new build of 6094b20851dcff871b6dc2a8953a2d818b65be2a has completed succesfully!
Binaries available at https://build.nav.community/binaries/patch-2

PR #660, opened 2020-02-17
See you in a hundred years

mxaddict:
See you in a hundred years

### DIFF
--- a/src/amount.h
+++ b/src/amount.h
@@ -28,7 +28,7 @@ extern const std::string CURRENCY_UNIT;
  * critical; in unusual circumstances like a(nother) overflow bug that allowed
  * for the creation of coins out of thin air modification could lead to a fork.
  * */
-static const CAmount MAX_MONEY = 60000000 * COIN;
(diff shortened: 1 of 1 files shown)


src/qt/styles/moc_dark.cpp
```diff
@@ -0,0 +1,89 @@
+/****************************************************************************
```
aguycalled: unrelated file

src/qt/styles/moc_light.cpp
```diff
@@ -0,0 +1,89 @@
+/****************************************************************************
```
aguycalled: unrelated file

src/qt/styles/moc_dark.cpp
mxaddict: Sorry, removed :)

mxaddict:
Capture
Patch works :)

navbuilder:
A new build of 1479a1a490c469e0297faf863c9b5a4ab77c0826 has completed succesfully!
Binaries available at https://build.nav.community/binaries/100years

PR #666, opened 2020-02-20
Updated qa/rpc-tests/cfunddb-statehash.py to use connect_nodes_bi

mxaddict:
I noticed this test would sometimes fail.

I debugging locally it seems that it would fail to reconnect the nodes after disconnecting them.

So I updated the test to use connect_nodes_bi instead of the one way connection command, seems to have fixed it.


navbuilder:
A new build of 6aae817d97afce800185233b8c6a22ffe070544f has completed succesfully!
Binaries available at https://build.nav.community/binaries/fixed-cfunddb-statehash

PR #667, opened 2020-02-20
Do not use stempool if devnet and no peers

aguycalled:
Fixes a bug for devnet where transactions never leave the stempool and therefore are never confirmed when using devnet with a 1-node-network topology.


navbuilder:
A new build of db0ac24a70b8755f61992fa95406d17dff58ac7e has completed succesfully!
Binaries available at https://build.nav.community/binaries/dandelion-devnet-fix

PR #671, opened 2020-02-25
Fix fork length detection

aguycalled:


mxaddict:
72 blocks at 10 minutes would be 12 hours

aguycalled:
> Makes sense, I'm guessing these values were inherited from BTC?

yup

chasingkirkjufell:
i see, makes sense.

mxaddict:
Ack, I've approved the PR.

PR #678, opened 2020-05-19
[DAO] Fix for last voting cycle of proposals and payment requests

aguycalled:
Proposals and Payment Requests run for one cycle more than the one specified by the respective consensus parameters. This is preferred to be kept like this and not fixed, to maintain backwards compatibility with the validation of old entries.

This PR fixes the max cycle count shown in gui and rpc responses and translates new proposals for consensus changes based on this logic.

### DIFF
--- a/src/consensus/dao.cpp
+++ b/src/consensus/dao.cpp
@@ -2740,8 +2740,7 @@ std::string CProposal::GetState(uint32_t currentTime, const CStateViewCache& vie
     }
     if(currentTime > 0 && IsExpired(currentTime, view)) {
         sFlags = "expired";
-        if(fState != DAOFlags::EXPIRED && !ExceededMaxVotingCycles(view))
-            // This branch only occurs when a proposal expires due to exceeding its nDeadline during a voting cycle, not due to exceeding max voting cycles
+        if(fState != DAOFlags::EXPIRED)
             sFlags += ", waiting for end of voting period";
     }
     if(fState == DAOFlags::PENDING_VOTING_PREQ) {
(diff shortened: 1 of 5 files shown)


navbuilder:
A new build of 770cd19d34f25d920e3266dd354038d2d4a79379 has completed succesfully!
Binaries available at https://build.nav.community/binaries/fix-cfund-max-cycles

chasingkirkjufell:
on voting cycle 7, rpc command still shows voting cycle 6 with a status of expired (should be pending?) and state 0
image

The rpc command will correctly show the state of 3 at the last block of cycle 7 (showing cycle 6)

navbuilder:
A new build of 07bc3d671e68ff788f1ad104933285f4a981dde9 has completed succesfully!
Binaries available at https://build.nav.community/binaries/fix-cfund-max-cycles

mxaddict:
What do I need to test on this PR?

mxaddict:
> What do I need to test on this PR?

Here is what I know so far:
Cycle 0 usually does not count as there would not be enough blocks
So it should have cycles 0-7 ?

chasingkirkjufell:
> > What do I need to test on this PR?
> 
> Here is what I know so far:
> Cycle 0 usually does not count as there would not be enough blocks
> So it should have cycles 0-7 ?

Yeah, this PR fixes the RPC command of getproposal and getpaymentrequest where the return of votingcycle was displaying wrong numbers and the status as well. @mxaddict

mxaddict:
> > > What do I need to test on this PR?
> > 
> > 
> > Here is what I know so far:
> > Cycle 0 usually does not count as there would not be enough blocks
> > So it should have cycles 0-7 ?
> 
> Yeah, this PR fixes the RPC command of getproposal and getpaymentrequest where the return of votingcycle was displaying wrong numbers and the status as well. @mxaddict

Ack, will test this on Ubuntu 19.10

PR #682, opened 2020-05-22
Optimize memory use for blockindex

aguycalled:
This pull request optimizes the memory storage of the block indexes.

Block indexes store for every block the block header and some information needed for the chain validation. They are permanently kept in memory to speed up the validation process. Part of this extra data are the votes of the community fund/dao. Even if the votes of a block were empty, they were still using some memory space. This means even the first blocks of the chain (when no community fund was active) were taking memory space as if they had votes. This effect has been exaggerated with the introduction of the dao upgrade.

After the patch, a different set of data structures are used. Only blocks which include votes of a particular type attach data to those structures, hence reducing considerably the memory use. Other unused fields like prevoutStake and nStakeTime are also removed from memory.

What to test:

- memory use is effectively reduced when using this branch
- stressor tests run succesfully
- chain sync tests

### DIFF
--- a/src/chain.h
+++ b/src/chain.h
@@ -222,20 +222,11 @@ public:
     int64_t nCFSupply;
     int64_t nCFLocked;
 
-    std::vector<std::pair<uint256, int>> vProposalVotes;
-    std::vector<std::pair<uint256, int>> vPaymentRequestVotes;
-    std::map<uint256, bool> mapSupport;
-    std::map<uint256, uint64_t> mapConsultationVotes;
-
     std::string strDZeel;
 
     unsigned int nFlags;  // ppcoin: block index flags
 
     uint64_t nStakeModifier; // hash modifier for proof-of-stake
-
-    // proof-of-stake specific fields
-    COutPoint prevoutStake;
-    unsigned int nStakeTime;
 
     arith_uint256 hashProof;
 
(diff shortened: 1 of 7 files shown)


mxaddict:
Took a while to read the Diff, but changes seem to make sense, will do a build and some tests on mainnet.

* [ ] sync from genesis
* [ ] sync from genesis (restart in the middle) 
* [ ] run a the stress test a few times

mxaddict:
Btw @aguycalled I saw some lines modified that only changed the whitespace, might make sense to add a linter of some sort to enforce some rules on whitespace, indentation and formatting.

navbuilder:
A new build of 67a7a797901dd485d371ff5ca07330042ee00e3c has completed succesfully!
Binaries available at https://build.nav.community/binaries/optimize-blockindex-memory

navbuilder:
A new build of f4177a6a9b21a9863fdaf39824c91144e29de639 has completed succesfully!
Binaries available at https://build.nav.community/binaries/optimize-blockindex-memory

mxaddict:
Old master build was running 1.75GB RAM on my navcoind node, this branch is using about 1.5GB
So it's a nice ~200MB decrease.

mxaddict:
Ran the stress script 2 times for ~400 blocks each time, worked fine with this branch.

PR #689, opened 2020-06-03
[RPC] Return empty dict instead of throwing exception for unknown staker

aguycalled:
This PR modifies the rpc command `getstakervote` to return an empty dict instead of throwing an exception when queried for an unknown staker.

This is for better integration in electrumx.


navbuilder:
A new build of 246b73660fa5e0ad404b220da170f3d1aa90992e has completed succesfully!
Binaries available at https://build.nav.community/binaries/aguycalled-patch-5

mxaddict:
Tested prebuilt binary on 18.04

PR #696, opened 2020-06-04
[QT] Updated price update timer to respect shutdown request

mxaddict:
This bug was reported by Goku over on discord.

master/5.0.0 version of the wallet would fail to close if shutdown was request at exactly the time the price data was updated.

This patches that up so the wallet will respect shutdown request.


navbuilder:
A new build of 15c56ebec5f96cce081d7a99809c97360cb603b9 has completed succesfully!
Binaries available at https://build.nav.community/binaries/updated-price-update-to-respect-shutdown

PR #702, opened 2020-06-05
[RPC][GUI] Added more progress indicators during wallet start up, also shows via…

### LINKED ISSUE #654: Add progress info to "Loading block index..." RPC message
It would be useful if we could add the completion rate at the end of the message "Loading block index ...". In the NEXT wallet, we can only display the message that the block index is being loaded when the wallet is initialized, and users have no idea how long the block index loading process will take.

mxaddict:
closes #654

I'm sure this will be helpful for people wanting CLI progress during wallet start up, and also makes the splash screen have more feedback to the user.

But not sure how much this slows down the start up as it's basically tracking progress for most of the heavy loops in wallet start up.

@aguycalled do you think we should add this?

It's just something that came to mind when I took a look @ pr #654

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -1297,6 +1297,7 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler, const std
     if (fServer)
     {
         uiInterface.InitMessage.connect(SetRPCWarmupStatus);
+        uiInterface.ShowProgress.connect(SetRPCWarmupStatusProgress);
         if (!AppInitServers(threadGroup))
             return InitError(_("Unable to start HTTP server. See debug log for details."));
     }
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -5882,23 +5882,36 @@ CBlockIndex * InsertBlockIndex(uint256 hash)
 
 bool static LoadBlockIndexDB()
 {
+    uiInterface.InitMessage(_("Loading block guts..."));
     const CChainParams& chainparams = Params();
     if (!pblocktree->LoadBlockIndexGuts(InsertBlockIndex, InsertProposalVotes, InsertPaymentRequestVotes, InsertSupport, InsertConsultationVotes))
         return false;
 
     boost::this_thread::interruption_point();
 
+    uiInterface.InitMessage(_("Loading block index..."));
+
+    int nMapBlockInc = 0;
+
     // Calculate nChainWork
     vector<pair<int, CBlockIndex*> > vSortedByHeight;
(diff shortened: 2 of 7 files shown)


mxaddict:
Ok, I did some testing with this code, the wallet only loads like ~1% to ~2% slower than without this change on my PC, can you guys test and let me know if it performance ok?

mxaddict:
Capture2
Capture
Capture3

Xa5r:
Rescanning (changing the wallet.dat) it also what to a very long time yesterday

navbuilder:
A new build of a455624bc3b1fe1eb8c16fc1d8a95c15156e4a61 has completed succesfully!
Binaries available at https://build.nav.community/binaries/added-more-progress-indicators

mxaddict:
> Rescanning (changing the wallet.dat) it also what took a very long time yesterday

@Xa5r could you test this PR and see if it makes your wallet start time slow?

Xa5r:
Its difficult to compare because the loading time is always different.
But I noticed that by bootstraping the notice "Loading block guts xxx..." goes over 4.3 million, more blocks than the current highest block.

mxaddict:
> Its difficult to compare because the loading time is always different.
> But I noticed that by bootstraping the notice "Loading block guts xxx..." goes over 4.3 million, more blocks than the current highest block.

I don't think that loading block guts only loads active chains, I think it also loads forked/abandoned chain heads, which might explain why it loads more than the active chain.

I'm just assuming though, might be better to ask @aguycalled about it.

aguycalled:
> > Its difficult to compare because the loading time is always different.
> > But I noticed that by bootstraping the notice "Loading block guts xxx..." goes over 4.3 million, more blocks than the current highest block.
> 
> I don't think that loading block guts only loads active chains, I think it also loads forked/abandoned chain heads, which might explain why it loads more than the active chain.
> 
> I'm just assuming though, might be better to ask @aguycalled about it.

yes that's right

aguycalled:
tested gitian osx, surprisingly, my wallet in mainnet loads 10 times faster with this branch

mxaddict:
> tested gitian osx, surprisingly, my wallet in mainnet loads 10 times faster with this branch

I don't know how that would happen, but it's a good thing I guess 🗡️

mxaddict:
> > tested gitian osx, surprisingly, my wallet in mainnet loads 10 times faster with this branch
> 
> I don't know how that would happen, but it's a good thing I guess 🗡️

@chasingkirkjufell can you maybe test this branch as well on windows?

I've tested on Ubuntu 18.04 (Which performed about 1-2% slower)

PR #715, opened 2020-06-12
[TEST] Fixed qa/rpc-tests/cfunddb-statehash.py

mxaddict:
Test was failing due to the nodes not finding the best chain

Updated the test to create more blocks in between chain syncs so it can find the best chain faster


navbuilder:
A new build of 205c4f4e61815c394094c109eeee9b9563b74a63 has completed succesfully!
Binaries available at https://build.nav.community/binaries/fixed-cfunddb-statehash.py

PR #730, opened 2020-07-06
[WALLET] Fix -wallet issue #729

### LINKED ISSUE #729: [GUI] config variables wrong precedence/order on wallet startup
Describe the issue:
when starting up the navcoin wallet 5.0.1 with e.g. a 'wallet=mywallet.dat' directive specified in custom config file ('navcoin.conf')
the 'mnemonic startup gui' is presented, although the valid 'mywallet.dat' file is present.
at that point i don't want to create a new wallet, because i have already one....
the 'mnemonic startup gui' vanishes, when the custom wallet file is renamed  to the default 'wallet.dat' name.

i assume the startup logic is flawed, in that the presence of the default file is checked, but not the change of the default setting.
here, the custom configuration should take precedence over the internal set values,
because what sense do config variables for changing default values make, when they are not recognized ?  

What behavior did you expect? 
the 'mnemonic startup gui' should recognize custom config variables of the configuration file or cli specified options.
i.e. check not only default settings but also changes to these settings...
 
How reliably can you reproduce the issue, what are the steps to do so?
can easily reproduced by renaming the default 'wallet.dat' file to custom name, adjusting that value in configuration file and then startup wallet.

What version of NavCoin Core are you using, where did you get it (website, self-compiled, etc)?
NavCoin 5.0.1 from github release repository
windows 7 64 bit

aguycalled:
Fixes #729

### DIFF
--- a/src/qt/navcoin.cpp
+++ b/src/qt/navcoin.cpp
@@ -435,9 +435,6 @@ bool NavCoinApplication::setupMnemonicWords(std::string& wordlist, std::string& 
         return true;
     }
 
-    std::string walletFile = GetArg("-wallet", "wallet.dat");
-    if (fs::exists(walletFile)) return true;
-
 	if (CheckIfWalletDatExists()) return true;
 
     StartOptionsMain dlg(nullptr);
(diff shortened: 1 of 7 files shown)


navbuilder:
A new build of 8b3413e8951bf94945e344dfb7feb42c868eea60 has completed succesfully!
Binaries available at https://build.nav.community/binaries/fix-wallet-issue-729

mxaddict:
Tested QT Wallet Ubuntu 18.04 gitian build

PR #732, opened 2020-07-15
[RPC] getaddresshistory counts balance for whole history even when range

aguycalled:


navbuilder:
A new build of 948936eb82bbaee3100b3ae903103eef0e39a178 has completed succesfully!
Binaries available at https://build.nav.community/binaries/aguycalled-patch-8

PR #745, opened 2020-10-11
[DAO] Exclude inactive stakers from votings

aguycalled:
This PR proposes a consensus change signaled by version bit 28 to allow excluding stakers from the DAO votings.

Nodes (like users who do not want to participate in the votings or exchanges which activate staking) will be able to exclude their staked blocks from the quorum by indicating the option `-excludevote=1`, preventing a scenario where a big enough amount of staking coins not engaging in the governance process could make impossible to reach the acceptance/rejection thresholds.

Additionally, a node will label automatically their blocks to be excluded even if `-excludevote=1` is no specified, whenever the staker hasn't been active voting in any of the last 10 voting cycles which had an active vote. This is reverted whenever the staker casts a vote.

Blocks are labeled to be excluded by activating the right-most bit of their `nOnce` parameter.


mxaddict:
untest ACK

navbuilder:
A new build of f8ff873d14871e0e941546288c239af7e4e26d7d has completed succesfully!
Binaries available at https://build.nav.community/binaries/exclude

navbuilder:
A new build of 737812dc34d754769b67800010a9204a8cf10ac5 has completed succesfully!
Binaries available at https://build.nav.community/binaries/blsct

chasingkirkjufell:
737812d is a blsct commit. the build bot posted to the wrong thread.

chasingkirkjufell:
The text for showing voting status is not accurate. In this example, I have generate 5 blocks with "excludevote" node so the voting cycle block count should is 5. I voted1yes with 4 blocks remaining. Wallet thinks that it already passes both the quorum and amount of yes votes needed. The bug is not impactful due to the fact as voting continues, the wallet will eventually update to correct quorum and behavior.
image

aguycalled:
should be fixed now. i also added a rpc command `setexclude` to change on the fly the node exclude status

navbuilder:
A new build of a6ac923aee79b3167547fc7efccb630dca039b83 has completed succesfully!
Binaries available at https://build.nav.community/binaries/exclude

chasingkirkjufell:
I couldn't get the node to switch to exclude mode even after 10 cycles of inactivity. Is there a way to check if the mode has entered exclude mode?

aguycalled:
were votes happening during those 10 cycles?

chasingkirkjufell:
Does other nodes need to be voting? ie. if the proposal is not getting any votes, will it be counted into the exclude cycle?

src/consensus/dao.cpp
```diff
@@ -2645,7 +2645,7 @@ bool CPaymentRequest::IsAccepted(const CStateViewCache& view) const
     if (nVersion & ABSTAIN_VOTE_VERSION)
         nTotalVotes += nVotesAbs;
 
-    return nTotalVotes > (GetConsensusParameter(Consensus::CONSENSUS_PARAM_VOTING_CYCLE_LENGTH, view) * nMinimumQuorum) - exclude
+    return nTotalVotes > ((GetConsensusParameter(Consensus::CONSENSUS_PARAM_VOTING_CYCLE_LENGTH, view) - exclude) * nMinimumQuorum)
             && ((float)nVotesYes > ((float)(nTotalVotes) * GetConsensusParameter(Consensus::CONSENSUS_PARAM_PAYMENT_REQUEST_MIN_ACCEPT, view) / 10000.0));
```
chasingkirkjufell, line 2649: The second check currently could be reverted if no votes are increasing all of a sudden. Shouldn't the second check be something like this? 
`nVoteYes > (nTotalVotes + remaining blocks in the cycle) * GetConsensusParameter(Consensus::CONSENSUS_PARAM_PAYMENT_REQUEST_MIN_ACCEPT, view) / 10000.0)`

aguycalled:
> Does other nodes need to be voting? ie. if the proposal is not getting any votes, will it be counted into the exclude cycle?

if no one votes on a cycle, then this cycle does not count in the auto exclude counter

src/consensus/dao.cpp
aguycalled: id say because of the first check (nTotalVotes is larger than the min quorum), nTotalVotes will always have a value in a valid range so i don't see the need to add the remaining blocks in the cycle. but maybe i'm understanding it wrong?

chasingkirkjufell: The current checks are correct logically. The only thing that came to my mind is that there will be cases where the checks pass and the text are displayed like "accepted, waiting for end of voting cycle". It is not technically correct since there is still a chance where the no votes suddenly increase and overthrow the decision. 
For example,
If block cycle has 10 blocks, first 5 are yeses, then the checks will pass and the text will be displayed. However, there is still a chance where the remaining 5 could overthrow that outcome. In this scenario, the proposed checks will not pass so it'd still be pending.

All in all, I'd say the proposed check may not be needed if we just change the displayed text to show something different than "Accepted....." since it's not accepted yet.

aguycalled: mmm i would say the message is only shown when there are 50%+1 votes (because of the > sign) and therefor it cant be reversed

chasingkirkjufell: it can since the required yes votes are 70%. So cycle of 10 blocks, first 6 blocks are yes, at this stage it passes both checks (pass 50% quorum and 70% yes) so it'd show accepted. But if the remaining 4 blocks are all no votes, then we would end up having only 60% approval and the accepted is reverted right?

aguycalled: ok now i got it. you are right

aguycalled: so you mean something more like this:
```c++
(float)nVotesYes > (float)(GetConsensusParameter(Consensus::CONSENSUS_PARAM_VOTING_CYCLE_LENGTH, view) - exclude) * GetConsensusParameter(Consensus::CONSENSUS_PARAM_PAYMENT_REQUEST_MIN_ACCEPT, view) / 10000.0
```

chasingkirkjufell: not exactly, this will result in a static amount of yes votes needed while the proposed one is more dynamic basing of current voting conditions. 
For example, with this check, assuming there are no exclude votes, required yes votes are always 7 in the case of 10 block cycle. The proposed one could have much less than 7 votes. 
Say the first 5 votes are no votes (not abstain and not exclude but just not voting), and adding 4 yes votes, the proposed check will pass while this check won't pass.

aguycalled: the problem is that this function can't see what is the current block inside of the cycle so it can't know how many blocks are left

chasingkirkjufell: i see. maybe we can change the text then? Like to "en-route to acceptance" for something in those lines?

aguycalled: what about just "pending" when the cycle did not end

chasingkirkjufell: yes that works as well. is the text using the same checks here? How does it work when these checks changed from false to true but it's pending all the way?

aguycalled: it would check for a state change which would only happen after the end of a cycle

chasingkirkjufell: sounds good.

navbuilder:
A new build of 8489cee059de39b62e71e65c2a88769bcd798eee has completed succesfully!
Binaries available at https://build.nav.community/binaries/exclude

chasingkirkjufell (approved):
Tested local build on Ubuntu 20.04

- Wallet launches with -excludevote=1
- Votes are not counted when wallet is launched with excludevote=1 for proposal consultations, and consensus changes
- Proposal and consultation quorum is reduced when blocks are mined by excludevote nodes
- Wallet enters excludevote mode after 10 cycles of inactivity

mxaddict:
@aguycalled @chasingkirkjufell how do I know if a node has entered into `excludevote mode`?

chasingkirkjufell:
i don't believe you can @mxaddict

chasingkirkjufell:
i verified it by generating blocks and seeing that quorum has dropped with another node voting.

mxaddict:
Gotcha

aguycalled:
the mined blocks would have nonce==1

navbuilder:
A new build of ae49b2ec3e29303453b3c5123ba13fd0ee42053d has completed succesfully!
Binaries available at https://build.nav.community/binaries/exclude

mxaddict:
@aguycalled can you explain a bit more about the changes to the tests? I noticed that the status checks have all been updated to just check == "pending" in most of the test scenarios.

mxaddict:
I assume this change is only for tests that were expecting a variant of `*waiting for end of voting period`.

aguycalled:
yes, you  are right

mxaddict:
> yes, you are right

Ok, cool.

I just checked the source that was generating the text and now it makes sense.

mxaddict (approved):

PR #778, opened 2021-02-02
Apply -blsctmix to threads and gui/rpc send

aguycalled:
This PR takes in account if blsctmix has been turned off (blsctmix=0) to disable the background threads and aggregation when sending xNAV.

What to test:

With blsctmix=0 there is no aggregation for xNAV transactions, but there is when it is turned on.


0x2830 (approved):
Approved

mxaddict (approved):

PR #790, opened 2021-02-18
Fixed a reference to BLS

mxaddict:


aguycalled (approved):

chasingkirkjufell:
What does this mean for the current master that doesn't have this fix?

mxaddict:
> What does this mean for the current master that doesn't have this fix?

Yes, but this is only for the bench mark binary, which is not released during builds, so current release is fine.

chasingkirkjufell (approved):

PR #798, opened 2021-02-20
Prevent adding duplicated inputs from candidates

aguycalled:

### DIFF
--- a/src/blsct/aggregationsession.cpp
+++ b/src/blsct/aggregationsession.cpp
@@ -179,6 +179,8 @@ bool AggregationSession::SelectCandidates(CandidateTransaction &ret)
     unsigned int i = 0;
     unsigned int nSelected = 0;
 
+    std::set<uint256> setSeen;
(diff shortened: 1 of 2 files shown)


mxaddict:
Ack, compiling, how to test this change?

aguycalled:
i don't think there's a proper way to test it, i just seen sometimes the same previous outputs is considered twice as a candidate, making the candidate aggregation fail.
the pr fixes by checking that each previous output has not been added before

mxaddict:
> i don't think there's a proper way to test it, i just seen sometimes the same previous outputs are considered twice as a candidate, making the candidate aggregation fail.
> the pr fixes this by checking that each previous output has not been added already

Makes sense, I guess just a smoke test with a swap and send xnav -> xnav should be good enough to test it.

mxaddict (approved):

chasingkirkjufell (approved):

PR #801, opened 2021-02-24
GUI lock up optimization

mxaddict:
Optimized how the wallet manages GUI updates from wallet tx data and reports for balance and stake report


navbuilder:
A new build of 43a79c992400060ca94a331a300461394236c745 has completed succesfully!
Binaries available at https://build.nav.community/binaries/gui-lock-up-optimization

navbuilder:
A new build of d393fe2f343be41ccc3dd86e556c624f01fb54b9 has completed succesfully!
Binaries available at https://build.nav.community/binaries/gui-lock-up-optimization

src/blsct/aggregationsession.cpp
```diff
@@ -951,6 +951,7 @@ void AggregationSessionThread()
                 MilliSleep(1000);
             } while (true);
 
+            MilliSleep(5000);
```
aguycalled, line 954: should we maybe let the user specify the timeouts so they can try with different settings depending on their specs?

mxaddict:
Yeah, we can make it a CLI param, and GUI setting.

navbuilder:
A new build of 57bd9830ec645ae94c069c6513d55de79b2a2dcd has completed succesfully!
Binaries available at https://build.nav.community/binaries/gui-lock-up-optimization

chasingkirkjufell:
this helps with the lock a lot. typically a 3 second lock when loading a new block is now about 1 second with some micro freezes now and then. 

Do you want to make sleep time configurable or this PR is ready as is?

mxaddict:
> this helps with the lock a lot. typically a 3 second lock when loading a new block is now about 1 second with some micro freezes now and then.
> 
> Do you want to make sleep time configurable or this PR is ready as is?

I'm adding the timeouts/sleeps as cli params, I'll push that change now.

mxaddict:
@aguycalled  @chasingkirkjufell I've added the CLI args to allow for users to change the blsct aggregation and verification thread sleep times.

mxaddict:
I tested it by adding a log for the value of `GetRand()` return value, and got these:

image

src/random.cpp
```diff
@@ -150,19 +150,33 @@ void GetStrongRandBytes(unsigned char* out, int num)
     memory_cleanse(buf, 64);
 }
 
-uint64_t GetRand(uint64_t nMax)
+uint64_t GetRand(uint64_t nMin, uint64_t nMax)
 {
     if (nMax == 0)
         return 0;
 
+    if (nMin == 0)
```
aguycalled, line 158: here 0 is returned if nMin equals 0

mxaddict:
Ahh, i see it now.

mxaddict:
@aguycalled  fixed the issue with GetRand, I was not seing it for some reason, the code was right in front of me 😃

navbuilder:
A new build of ea3f2c484ea2bbf39b735b3d84acea47263bb0b4 has completed succesfully!
Binaries available at https://build.nav.community/binaries/gui-lock-up-optimization

chasingkirkjufell (approved):
wallet runs a lot better with this PR, when a new block is found, the freeze time decreased from 2-3 seconds to about 1 second. However, micro freezes are more often with this PR. Adjusting the sleep time doesn't seem to play a big role in overall wallet performance visually.

mxaddict:
@aguycalled  anything else we can add to this branch before merging?

aguycalled (approved):
tested osx

PR #805, opened 2021-03-02
Prevent duplicate remove vote

aguycalled:
When a staker removed a vote for a proposal or payment request, it keeps broadcasting the remove vote in every block instead of using the cache and broadcasting it only in one block.

Example: https://www.navexplorer.com/block/4987836

This PR fixes the behavior.


navbuilder:
A new build of 71dcc27204510789f6b00da4deb784d53f0db36a has completed succesfully!
Binaries available at https://build.nav.community/binaries/prevent-duplicate-remove-vote

mxaddict (approved):

chasingkirkjufell (approved):
tested gitian build in Windows 10

PR #854, opened 2021-10-04
Fix typo

aguycalled:
Fixes a typo which prevented to correctly parse "true" as a boolean when passed to the `viewaggregationsession` RPC command.

## What to test

RPC command `viewaggregationsession true` works.


chasingkirkjufell:
Started a review pointing out a missing a but it's not shown here.

PR #855, opened 2021-10-08
Protocol Upgrade: Burn Fees

aguycalled:
This PR introduces a consensus change proposal signalled by version bit 9.

If the change gets approved, stakers would not get the transactions fees of their mined blocks added to the reward. This would effectively burn the transaction fees.

### What to test

Once the consensus change is approved, blocks including transactions which pay a fee greater than 0, only pay the staking reward to the address which found the block.


navbuilder:
A new build of ff85698e041100812847326f322eb5f120da4d72 has completed succesfully!
Binaries available at https://build.nav.community/binaries/burn_fees

ML-NAV:
Would be good to test if the navexplorer works fine after this change. Just to be sure/safe.

PR #857, opened 2021-10-16
Updated to work with c++14 standard instead of c++11

mxaddict:


mxaddict:
I think for this change we just need to make sure that the gitian builds are working correctly, nothing else was changed in this PR

mxaddict:
Closing in favor of #861

PR #862, opened 2021-10-18
xNAV upgrades for Navcoin Core 7.0.0

aguycalled:
This Pull Request introduces various consensus changes related to xNAV signalled by version bit 33. As we did run out of version bits for updates, this PR also introduces a patch to use the nBits field of the block header to signal for version bits greater than 32.

As the right-most bit of nBits is already used to signal blocks excluded from voting, there are 31 bits left for version bit signalling. Therefor the max version bit supported at the moment is 63 and `nBit` version signalling starts at the second right-most bit of the field.

### Improvements in the xNAV inter-node mixing protocol.

Sending the public key associated to a candidate transaction leaks no private details about it. By broadcasting the candidate transaction together with the candidates, allows node to filter which transaction they should try to decrypt instead of trying to decrypt every transaction, hence greatly improving CPU load.

### Optimizations in the xNAV verification routines.

There have been improvements in the verification functions of the cryptographic proofs attached to xNAV transactions, resulting in huge speed increases. Seeing numbers down to 15ms in my laptop.

### Changes in the serialisation format of outputs.

- Transaction outputs with the new serialisation format optimise space by using the `value` field to signal which fields contain data and hence only (de)serialising those. (Previous situation would (de)serialise all the fields whether they had useful data or not).

For example, the amount of an output is never assigned space when the output is private. Before it was an 8 byte field filled with zeroes for private transactions. Similar situation was given for NAV outputs in a transaction with xNAV inputs, where the spendingKey and ephemeralKey were not significant but were still consuming 96 bytes of space in total until this patch is applied.

- Two new fields are optionally added to outputs: `std::vector<unsigned char> vData` and `uint256 tokenId`. They will only (de)serialise when their values differ from their initial state.

`vData` can be used to hold arbitrary data. The original intent of this is to substitute the use of the transaction's field `strDZeel` to hold metadata or program predicates. As the `strDZeel` fields of different xNAV transactions can't be aggregated into one, we were missing the extended use derived from the utility of `strDZeel` for private transactions.

`tokenId` signals the token a private output refers to. Its default value `0x0` refers to xNAV. From `tokenId` it is calculated the generator `H` used in the verification of the Bulletproof Rangeproofs without breaking the zero-sum-balance property in the xNAV verification protocol.

### Confidential Tokens: nAssets

This Pull Request introduces Confidential Tokens. They are disabled initially in Mainnet,  but allowing its late activation when the network decides to, through a consensus change vote without the need of a hard fork. Testnet, Devnet and Regnet will have it activated as to allow testing of the new feature.

Built on top of the same cryptographic protocol which makes xNAV possible, we are adding the possibility to mint confidential tokens on the Navcoin blockchain. Transactions of those tokens will veil the details of their sender, receiver and amount transferred.

nAssets can represent arbitrary tokens created by users of the Navcoin network as well as stablecoins with their value pegged and backed by different underlying assets, while benefiting from the privacy-preserving properties of blsCT.

### RPC Commands

- `createtoken <name> <symbol> <max_supply>`

Creates a Confidential Token with name `name`, symbol `symbol` and max supply `max_supply`. If `max_supply` is not indicated, its value will be 100,000,000,000 with divisions down to 8 decimal precision. Names can be duplicated and therefor the token id should be used to differentiate them.

xNAV is used to pay the creation fees.

The token will be managed by a BLS private key from the wallet which runs the RPC command. Token Keys are derived from the master BLS key using the `m/130'/2'/index` path.

- `listtokens`

Shows a list of network-wide available tokens together with their properties and wallet balance of that token, if any.

- `minttoken <token_id> <xnav_address> <amount>`

Mints `amount` of tokens to `xnav_address`. Minting tokens add to the token's current supply and can be executed until the max. supply of a token is depleted.

xNAV is used to pay the minting fees.

- `sendtoken <token_id> <xnav_address> <amount>`

Sends `amount` of tokens to `xnav_address` from current wallet.

xNAV is used to paid the transaction fees.

- `burntoken <token_id> <amount>`

Burns `amount` of tokens from current wallet.

xNAV is used to paid the transaction fees.

## What to test

- xNAV mixing on mainnet and testnet.
- xNAV transaction sending on mainnet and testnet.
- Mainnet syncing, confirm speed improvement.
- Test Confidential Tokens in devnet. Ideally through an stressing script.


navbuilder:
A new build of 8dfea9f55e3189ee5584eb603a0ef0e00296a0ed has completed succesfully!
Binaries available at https://build.nav.community/binaries/xnav-700

navbuilder:
A new build of c4471e364990543fd8da0300cdd3c38a1e7aa69d has completed succesfully!
Binaries available at https://build.nav.community/binaries/xnav-700

navbuilder:
A new build of 823788c2f3f828a1cbfe305ca8918e985bc8f9cc has completed succesfully!
Binaries available at https://build.nav.community/binaries/xnav-700

navbuilder:
A new build of be953aa11a58ed18fea9a05fc9fecc4904a8eb15 has completed succesfully!
Binaries available at https://build.nav.community/binaries/xnav-700

navbuilder:
A new build of fae2c5d13a2e948498b6f5cbded7e90c101bb990 has completed succesfully!
Binaries available at https://build.nav.community/binaries/xnav-700

chasingkirkjufell (approved):
created a stressor for the private tokens and stressed it with issues

PR #890, opened 2021-11-25
Updated c++std for bdb, boost, qt and zeromq

mxaddict:


chasingkirkjufell:
is gitian failing for this one?

mxaddict:
I think so, will check it out later today

mxaddict:
I think the build for this is failing based on the name of the branch, gitian is not liking the ++ symbols in the branch name, I'll re-create the PR with a different branch name

PR #898, opened 2021-12-04
Fix ExtractDestination() for ColdStaking

aguycalled:
What to test:

* Cold Staking works correctly with this patch

### DIFF
--- a/src/script/script.cpp
+++ b/src/script/script.cpp
@@ -231,6 +231,71 @@ bool CScript::IsColdStakingv2() const
             (*this)[23] == OP_IF);
 }
 
+bool CScript::HasColdStakingP2PKHStaking() const {
+    return (this->size() > 27 &&
+            (*this)[0] == OP_COINSTAKE &&
+            (*this)[1] == OP_IF &&
+            (*this)[2] == OP_DUP &&
+            (*this)[3] == OP_HASH160 &&
+            (*this)[4] == 0x14 &&
+            (*this)[25] == OP_EQUALVERIFY &&
+            (*this)[26] == OP_CHECKSIG &&
+            (*this)[27] == OP_ELSE);
+};
+
+bool CScript::HasColdStakingP2PKHSpending() const {
+    return (this->size() == 54 &&
+            (*this)[0] == OP_COINSTAKE &&
+            (*this)[1] == OP_IF &&
+            (*this)[2] == OP_DUP &&
+            (*this)[3] == OP_HASH160 &&
+            (*this)[4] == 0x14 &&
+            (*this)[25] == OP_EQUALVERIFY &&
+            (*this)[26] == OP_CHECKSIG &&
+            (*this)[27] == OP_ELSE &&
+            (*this)[28] == OP_DUP &&
+            (*this)[29] == OP_HASH160 &&
+            (*this)[30] == 0x14 &&
+            (*this)[51] == OP_EQUALVERIFY &&
+            (*this)[52] == OP_CHECKSIG &&
+            (*this)[53] == OP_ENDIF);
+};
+
+bool CScript::HasColdStakingv2P2PKHStaking() const {
+    return (this->size() > 50 &&
+            (*this)[0] == 0x14 &&
+            (*this)[21] == OP_DROP &&
+            (*this)[22] == OP_COINSTAKE &&
+            (*this)[23] == OP_IF &&
+            (*this)[24] == OP_DUP &&
+            (*this)[25] == OP_HASH160 &&
+            (*this)[26] == 0x14 &&
+            (*this)[47] == OP_EQUALVERIFY &&
+            (*this)[48] == OP_CHECKSIG &&
+            (*this)[49] == OP_ELSE);
+
+};
+
+bool CScript::HasColdStakingv2P2PKHSpending() const {
+    return (this->size() == 76 &&
+            (*this)[0] == 0x14 &&
+            (*this)[21] == OP_DROP &&
+            (*this)[22] == OP_COINSTAKE &&
+            (*this)[23] == OP_IF &&
+            (*this)[24] == OP_DUP &&
+            (*this)[25] == OP_HASH160 &&
+            (*this)[26] == 0x14 &&
+            (*this)[47] == OP_EQUALVERIFY &&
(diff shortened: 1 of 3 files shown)


mxaddict:
Building with Ubuntu 20.04

mxaddict:
Testing the cold staking now:

image


Will let this run for a while, will approve once I have had the cold staking running for about an hours

aguycalled:
Good! It should also get tested that spendable and stakable balances are correctly shown, my tests were successful on that regard.

mxaddict:
> Good! It should also get tested that spendable and stakable balances are correctly shown, my tests were successful on that regard.

So far tha balances are correct on the UI as well, but will monitor it as it keeps staking.

mxaddict:
@aguycalled "Pending" balance is only for when a TX adds balance but is not yet staked in a block correct?

aguycalled:
yes

mxaddict:
I've setup 2 cold staking nodes and 1 spending node, seems to be working nicely so far, will let these run for a while longer:

image

mxaddict (approved):
I think I'm happy with the testing on this PR, seems all is working with regards to cold staking, also testing some hot wallet staking and was still working as expected.

src/script/script.cpp
```diff
@@ -231,6 +231,71 @@ bool CScript::IsColdStakingv2() const
...
+            (*this)[28] == OP_DUP &&
+            (*this)[29] == OP_HASH160 &&
+            (*this)[30] == 0x14 &&
+            (*this)[51] == OP_EQUALVERIFY &&
+            (*this)[52] == OP_CHECKSIG &&
+            (*this)[53] == OP_ENDIF);
+};
+
+bool CScript::HasColdStakingv2P2PKHStaking() const {
+    return (this->size() > 50 &&
```
RealKeyboardWarrior, line 265: must be ==

src/script/script.cpp
```diff
@@ -231,6 +231,71 @@ bool CScript::IsColdStakingv2() const
             (*this)[23] == OP_IF);
 }
 
+bool CScript::HasColdStakingP2PKHStaking() const {
+    return (this->size() > 27 &&
```
RealKeyboardWarrior, line 235: must be ==

RealKeyboardWarrior (changes requested):

src/script/script.cpp
```diff
@@ -231,6 +231,71 @@ bool CScript::IsColdStakingv2() const
...
+            (*this)[0] == OP_COINSTAKE &&
+            (*this)[1] == OP_IF &&
+            (*this)[2] == OP_DUP &&
+            (*this)[3] == OP_HASH160 &&
+            (*this)[4] == 0x14 &&
+            (*this)[25] == OP_EQUALVERIFY &&
+            (*this)[26] == OP_CHECKSIG &&
+            (*this)[27] == OP_ELSE);
```
RealKeyboardWarrior, line 243: why is it not validated beyond this point?

src/script/script.cpp
```diff
@@ -231,6 +231,71 @@ bool CScript::IsColdStakingv2() const
...
+            (*this)[0] == 0x14 &&
+            (*this)[21] == OP_DROP &&
+            (*this)[22] == OP_COINSTAKE &&
+            (*this)[23] == OP_IF &&
+            (*this)[24] == OP_DUP &&
+            (*this)[25] == OP_HASH160 &&
+            (*this)[26] == 0x14 &&
+            (*this)[47] == OP_EQUALVERIFY &&
+            (*this)[48] == OP_CHECKSIG &&
+            (*this)[49] == OP_ELSE);
```
RealKeyboardWarrior, line 275: ditto why is it not validated beyond this point?

RealKeyboardWarrior (changes requested):

src/script/script.cpp
RealKeyboardWarrior: nvm - it's just to get the coldstake key

src/script/script.cpp
RealKeyboardWarrior: nvm - it's just to get the coldstake key

src/script/script.cpp
RealKeyboardWarrior: nvm - it's just to get the coldstake key

src/script/script.cpp
RealKeyboardWarrior: nvm - it's just to get the coldstake key

chasingkirkjufell:
building with ubuntu 20.04 as well, will report any findings

chasingkirkjufell:
when i try to send coins to the cold staking address the balance isn't correct.
steps:
create 2 wallets, create a coldstaking address v1 with wallet 1 benign spending and wallet 2 being staking
disable staking for both wallets to have better control
send multiple transaction from wallet 1 to the cold staking address
generate a block
image

chasingkirkjufell:
so it looks like the staking wallet is showing the correct balance since actually there are only 2 groupings that went through? history shows all 5 though. window on the left is from spending wallet while on the right is the staking wallet
image

mxaddict:
@chasingkirkjufell I did not run into any balance issues with my testing, can you give me steps to replicate?

aguycalled:
image
 
I haven't been able to replicate it. Most likely case is that the first transactions where spent in the following (in case they were sent to a cold staking address with a spending address of that node). Will wait to hear more details to see if we can replicate/confirm.

chasingkirkjufell:
> image
> 
> I haven't been able to replicate it. Most likely case is that the first transactions were spent in the following (in case they were sent to a cold staking address with a spending address of that node). Will wait to hear more details to see if we can replicate/confirm.

I noticed that this issue is not related to this PR so I am going to approve this PR for now. Lke your said, it may be because of the outputs were spent again but the transaction history is confusing. For what it's worth, here are a bit more detailed steps.

1. creating 2 wallets and disable staking
2. connect 2nd wallet to first
3. generate 300 blocks on 1st wallet
4. create a coldstaking address v1 with wallet 1 spending and wallet 2 staking
5. send 5 txs to coldstaking address
6. generate more blocks
7. the tab will show all 5 txes but only 2 are actually valid. repair wallet or zapwallettxes does not fix it

chasingkirkjufell (approved):

PR #900, opened 2021-12-07
Fix possible node crash if the stake was stale

aguycalled:
This pull request fixes a bug which can cause a node to crash when the stake is stale. This can be easily tested running various instances of a node staking with the same wallet.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -11140,6 +11140,8 @@ bool CheckProofOfStake(CBlockIndex* pindexPrev, const CTransaction& tx, unsigned
 
     if (fColdStaking)
     {
+        if (!view.HaveInputs(tx)) return error("%s: Coin stake %s is stale\n", __func__, tx.GetHash().ToString());
+
         CAmount valueIn = view.GetValueIn(tx);
         CAmount valueOut = 0;
 


mxaddict:
Away from my main PC right now, will wait for gitian build to test

aguycalled:
https://build.nav.community/binaries/aguycalled%3Afix-node-crash-stale-stake/

navbuilder:
A new build of 29e68570608e739c19c6662afda5770170018211 has completed succesfully!
Binaries available at https://build.nav.community/binaries/fix-node-crash-stale-stake

mxaddict:
Testing as follows:
Created a devnet wallet.dat
Imported into 5 nodes,
Have the 5 nodes staking at the same time.

mxaddict:
@aguycalled  would this testing method be a good way to determine if it's working?

aguycalled:
yep

mxaddict:
Ack

mxaddict:
> 

Tested this scenario and seemed to be fine, for some reason some of the nodes would get orphaned blocks even when the other nodes minted the same inputs into the same block (I guess a slight difference in timestamp for the block caused this)

But I think it's fine, the nodes did not crash while I was running them.

mxaddict (approved):

chasingkirkjufell:
oops, completely missed this one. didn't see a ready for review tag so i thought it's still work in progress.

PR #912, opened 2022-01-20
[FIX] Generate random key if not viewing key available

aguycalled:
This pull request prevents a segfault crash when verifying a block which redirected its rewards to a xNAV address if the verifier node has an old wallet and does not have a view key.


chasingkirkjufell:
Ack

navbuilder:
A new build of 88ce1d314269c39166f1f88390c3d790c28b4c46 has completed succesfully!
Binaries available at https://build.nav.community/binaries/fix-segfault

PR #928, opened 2022-02-07
Improved the response logging for curl call to price info

mxaddict:
image

### DIFF
--- a/src/qt/navcoingui.cpp
+++ b/src/qt/navcoingui.cpp
@@ -2057,6 +2057,7 @@ void NavcoinGUI::updatePrice()
 
     std::thread pThread{[this]{
         std::string response;
+        std::string headers;
         try {
             CURL *curl;
             std::string url(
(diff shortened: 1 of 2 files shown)


aguycalled:
do we want to merge this or do we just want to get binaries from gitian so pclark can test it?

mxaddict:
I think we should merge it as it will help with further debugging in the future if anyone else get's issues with the price info.

Or we can just use the build to help debug for pclark.

I'll leave it up to you.

aguycalled:
how often would it print to the debug? it it's too often, then we should only debug when some flags are enabled in the config file (debug=prices, for example) if we wanted to merge

chasingkirkjufell (approved):

mxaddict:
@aguycalled I think we should merge this after the build, but I think it's ready for review

navbuilder:
A new build of cc7a99cd56af89a4332347b1fb3c896001a92272 has completed succesfully!
Binaries available at https://build.nav.community/binaries/improved-logging-for-price-info-response

aguycalled (approved):

PR #930, opened 2022-02-08
Added is_mine property to listtokens return value if token balance is…

### LINKED ISSUE #923: Add is_mine property to listtokens rpc method
Currently `listtokens `method returns all available tokens in the blockchain.

I think it is necessary to add the `is_mine` attribute to every token object so that we can understand which of these tokens belongs to the wallet owner.

Sample :
```
  {
    "version": 0,
    "is_mine":true,
    "id": "777896868f6b80710b42923a0ad51bf6ff73fb0cbd727ad2d617c4c338683fca",
    "pubkey": "99259540dc2fae3fd6c3205be105f202c8416fe87433833f810ac255c422d26a78c4d3e594949f0ec17e820265b15f9a",
    "name": "Baby NAV",
    "token_code": "BNAV",
    "current_supply": "200.00",
    "max_supply": "100000000.00",
    "balance": "0.00"
  },
```

mxaddict:
closes #923

### DIFF
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -6041,22 +6041,23 @@ UniValue listtokens(const UniValue& params, bool fHelp)
             }
 
             // Is this token ours?
-            bool fTokenIsMine = true;
+            bool fTokenIsMine = false;
 
-            if (!pwalletMain->HaveBLSCTTokenKey(it->second.key))
+            blsctKey pk;
+            if (!pwalletMain->GetBLSCTTokenKey(it->second.key, pk))
             {
                 blsctKey sk;
 
                 if (!pwalletMain->GetBLSCTSpendKey(sk))
                     throw JSONRPCError(RPC_TYPE_ERROR, "Wallet not available");
 
-                blsctKey pk = sk.PrivateChildHash(SerializeHash("nft/"+it->second.sName+it->second.sDesc));
-
-                if (pk.GetG1Element() != it->second.key)
-                    fTokenIsMine = false;
+                pk = sk.PrivateChildHash(SerializeHash("nft/"+it->second.sName+it->second.sDesc));
 
                 pwalletMain->AddBLSCTTokenKey(pk);
             }
+
+            if (pk.GetG1Element() == it->second.key)
+                fTokenIsMine = true;
 
             o.pushKV("is_mine", fTokenIsMine);
             if (!fMine || (fMine && fTokenIsMine))
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -5989,7 +5989,7 @@ UniValue listtokens(const UniValue& params, bool fHelp)
     if (fHelp)
         throw std::runtime_error(
                 "listtokens (mine)\n"
-                "\nList the confidential tokens. Set mine to true to show only tokens with balance.\n"
+                "\nList the confidential tokens. Set mine to true to show only tokens you own.\n"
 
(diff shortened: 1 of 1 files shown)


aguycalled:
i think what sakdeniz meant with this request was to filter the entries which are owned by the wallet (independently of the balance). you can check the logic at the minttoken or mintnft rpc commands to see how to check if a token is owned (line 1622 rpcwallet.cpp for example)

mxaddict:
I see, should the (mine) param for listtokens be updated as well? Cause that shows tokens that have balance, but might not be owned by the wallet.

aguycalled:
yes probably, those tokens can easily be filtered without the current mine just checking for balance > 0 client side

mxaddict:
> yes probably, those tokens can easily be filtered without the current mine just checking for balance > 0 client side

Alright

mxaddict:
@aguycalled I've updated the logic, does this look good?

aguycalled:
i would need to test it, but initially it looks to me like it would label as mine tokens which are not owned by the wallet

navbuilder:
A new build of 8ead11a00555b7e5cff2d2c9747141631a1e1537 has completed succesfully!
Binaries available at https://build.nav.community/binaries/add-is-mine-to-listtokens-rpc-command

navbuilder:
A new build of 2d7e4345f919e7b6a5fa87a7f63aa84d257cf153 has completed succesfully!
Binaries available at https://build.nav.community/binaries/add-is-mine-to-listtokens-rpc-command

mxaddict:
@aguycalled this seems to be working with my testing but @sakdeniz said he was having issues, I'm still trying to investigate.

@sakdeniz could you let me know what steps you took so I can replicate your issue with this.

aguycalled:
from what I read in the code

```
             // Is this token ours?
             bool fTokenIsMine = true;
```

isMine is by default yes

```
             if (!pwalletMain->HaveBLSCTTokenKey(it->second.key))
             {
```

this block runs only if the wallet does not have the key for the token

```
                 if (pk.GetG1Element() != it->second.key)
                     fTokenIsMine = false;
```

wallet compares the stored key with the assigned to the token.

in my opinion it should be false by default and set it to true if the stored key *equals* the assigned to the token

but giving this opinion just from reading the code

mxaddict:
> 

Does this code return true even if the spend key does not own the token/nft?
```cpp
pwalletMain->HaveBLSCTTokenKey(it->second.key)
```

mxaddict:
If it returns true when spendkey does not own the token or nft, I think we will need to update the logic in `mintnft` and `minttoken` as well.

mxaddict:
> @aguycalled this seems to be working with my testing but @sakdeniz said he was having issues, I'm still trying to investigate.
> 
> @sakdeniz could you let me know what steps you took so I can replicate your issue with this.

@sakdeniz I think you might have been running the first build (Which was only checking balance)

navbuilder:
A new build of a9cafefed1c22a87042b738c4eaebfb53d8efa12 has completed succesfully!
Binaries available at https://build.nav.community/binaries/add-is-mine-to-listtokens-rpc-command

chasingkirkjufell (approved):
tested gitian build on windows 10

chasingkirkjufell:
just wondering, why not add a filter for the rpc instead? like `listtokens mine` wouldn't that be more straight forward and the output will be clean instead of trying to go through all of them and find the one has is_mine: true

sakdeniz:
Tested with latest build, and it works fine.

sakdeniz (approved):

